PR #1, opened 2021-04-26
Initial files

michaeldboyd:
- added `sdks`, `demo`, `proto` to repository


cli/src/cli.yaml
```diff
@@ -0,0 +1,346 @@
...
+                - name:
+                    value_name: STRING
+                    help: name for organization
+                    takes_value: true
+                    required: true
+                - id:
+                    value_name: STRING
+                    help: id for organization
+                    takes_value: true
+                    required: true
```
MichaelEdwardBlack, line 325: Is it actually required to pass an ID when creating an organization? I could be wrong but I thought this was an optional field

tmarkovski: Good catch, I don't think it is

michaeldboyd: yay!~ our team is talking over PRs. go team

PR #2, opened 2021-04-28
Create rust.yml

michaeldboyd:


tmarkovski:
Looks like you can pass `ssh-key` to the agent, and it'll work with private repos
https://github.com/actions/checkout/issues/287

PR #22, opened 2021-05-24
Black/2021 05 18/node tests

MichaelEdwardBlack:
Passing integration tests for the wallet and provider services
image

The provider service InvitationStatus test does not actually run because it is not implemented on the backend yet


.github/workflows/node.yml
```diff
@@ -0,0 +1,33 @@
...
+      run: 
+        working-directory: ./node
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with: 
+          token: ${{ secrets.ACCESS_TOKEN }}
+          node-version: ${{ matrix.node-version }}
+        uses: actions/checkout@v1
```
tmarkovski: May be missing a `-` here as in `- uses: actions/checkout@v1`

node/package.json
```diff
@@ -3,22 +3,27 @@
...
   },
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@grpc/grpc-js": "^1.2.8",
-    "google-protobuf": "^3.15.1"
+    "@trinsic/okapi": "^1.0.0-preview.10",
+    "google-protobuf": "^3.17.0",
+    "grpc": "^1.24.4",
+    "pb-util": "^0.1.3"
```
tmarkovski: Is this package used?

tmarkovski (approved):

node/package.json
MichaelEdwardBlack: nope, it's not. I'm removing it now

tmarkovski:
Fixes https://github.com/trinsic-id/server/issues/38

PR #30, opened 2021-06-02
Simple C# Added to Walkthrough

michaeldboyd:
- clarify index
- added csharp and annotations
- added c# script

TODOs
- Make c# scripts use vaccination files
- make a runnable c# script


cli/src/proto/trinsic_services.rs
```diff
@@ -181,6 +181,30 @@ pub struct VerifyProofResponse {
     #[prost(bool, tag = "1")]
     pub valid: bool,
 }
+#[derive(Clone, PartialEq, ::prost::Message)]
```
michaeldboyd, line 184: IDK how this got included. LMK if it needs to be removed

docs/walkthrough.md
```diff
@@ -1,23 +1,28 @@
...
+Let's walk through
+a scenario where we create a wallet and use it to issue, present, and verify a credential. 
 
-## Install the SDK
-If you're using our Gitpod Environment, the CLI and all the language specific SDKs have already been installed in your environment.
 
-If you're running this locally, you'll need to first install [rust](https://www.rust-lang.org/tools/install). Once installed run `cargo install --path .` in the `cli` directory.
+## 0. Install Trinsic
+Before we start, make sure you have an SDK installed.
+=== "CLI"
```
michaeldboyd: I like not having to make the user open a new tab for the install instructions, so I made each of these snippets. As we write how-to guides for different languages, I think these will come in handy. We could still just link to the install page in the future, but it was trivial to import the snippets

.gitignore
```diff
@@ -343,4 +343,6 @@ Cargo.lock
 .DS_Store
 
 .ionide
+
+Trinsic.IntegrationTests
```
tmarkovski: What is this ignoring?

docs/walkthrough.md
tmarkovski: Can we use "Trinsic CLI" universally as the tab name? That way, it get's persisted and applied everywhere.

mkdocs.yml
```diff
@@ -42,14 +44,16 @@ theme:
           icon: fontawesome/solid/sun
           name: Switch to light mode
 markdown_extensions:
-  - pymdownx.highlight
+  - pymdownx.highlight:
+      linenums: true
```
tmarkovski: If this can be enabled on per code block bases, that's be better. A lot of our snippets are one liners, and the "1" everywhere looked unnecessary, especially in script blocks, where it makes less sense.

tmarkovski (approved):

docs/walkthrough.md
michaeldboyd: Isn't "Trinsic CLI" kind of redundant? Because it's implied they're on Trinsic's docs site, I felt CLI was better. Thoughts?

PR #47, opened 2021-06-15
SDK for Browser

tmarkovski:
Making PR on behalf of @MichaelEdwardBlack


node/test/ProviderService.test.js
```diff
@@ -28,7 +28,7 @@ const createProfile = async () => {
 }
 
 test("make an invitation", async t => {
-  let providerService = new TrinsicProviderService();
+  let providerService = new TrinsicProviderService("20.84.208.207:5000");
```
michaeldboyd: Can we remove hardcoded default IP from source and use an env var instead?

node/test/WalletService.test.js
```diff
@@ -8,7 +8,7 @@ const { GenerateKeyRequest } = require('@trinsic/okapi');
 const { Struct } = require('google-protobuf/google/protobuf/struct_pb');
 
 test("get provider configuration", async t => {
-    let service = new TrinsicWalletService();
+    let service = new TrinsicWalletService("20.84.208.207:5000");
```
michaeldboyd: use env var

node/test/WalletService.test.js
```diff
@@ -18,7 +18,7 @@ test("get provider configuration", async t => {
 });
 
 test("create wallet profile", async t => {
-    let service = new TrinsicWalletService();
+    let service = new TrinsicWalletService("20.84.208.207:5000");
```
michaeldboyd: use env var

node/test/WalletService.test.js
```diff
@@ -61,7 +61,7 @@ test("generate proof with Jcs", async t => {
 })
 
 test("Demo: create wallet, set profile, search records, issue credential", async t => {
-    let walletService = new TrinsicWalletService();
+    let walletService = new TrinsicWalletService("20.84.208.207:5000");
```
michaeldboyd: use env var

web/src/proto/AuthService_grpc_web_pb.js
```diff
@@ -0,0 +1,159 @@
+/**
```
michaeldboyd: Is it standard practice to include the autogenerated proto code and the compiled typescript js files within the sdk source code? Seems these might be better accessed as a build command in the README. Lot's of autogenerated files in our version control now.

michaeldboyd (changes requested):

node/test/ProviderService.test.js
michaeldboyd: ideally with a force commit to overwrite it from version history

fundthmcalculus:
I can fix that when I get home. You shouldn't commit auto generated files,
that clutters the history and creates confusion.

node/test/WalletService.test.js
fundthmcalculus: Can we use a configuration in lieu of an env var?

web/src/proto/AuthService_grpc_web_pb.js
fundthmcalculus: No we shouldn't. Let me remove that, but add a build step to generate them.

tmarkovski: Ideally we shouldn't, but it's not too terrible that we do. Generating them is easy using the `Generate-Proto.ps1` script, so that can just be added as npm script pre-build.
Generated proto files should also be ignored in .gitignore and generation script which will use `protoc` with a plugin, should run consistently on all platforms, not just CI. Otherwise, it's best to keep the generated files in for now.

michaeldboyd (approved):

PR #53, opened 2021-06-26
SDK-44: Initial setup

fundthmcalculus:


tmarkovski (approved):
@fundthmcalculus This is not a request for change, just a note. We started generating the proto files during the build process, so that they're not included in source control. For .NET this is done automatically by the msbuild tasks, for Node and Browser, we just have a script step that does this.

Approving this PR anyway.

fundthmcalculus:
> @fundthmcalculus This is not a request for change, just a note. We started generating the proto files during the build process, so that they're not included in source control. For .NET this is done automatically by the msbuild tasks, for Node and Browser, we just have a script step that does this.
> 
> Approving this PR anyway.

For .NET, that makes sense. For Python, due to the complexity of getting `protoc` onto the CI/CD container and some post-build tweaks, it isn't worth it right now. Long-term, I do plan to do that. But that is predicated on the Python protobuf generation becoming more stable...

PR #58, opened 2021-06-28
Support different JSON formats in the PB messages

tmarkovski:
Proposed approach to use different formats for JSON other than Struct.
A number of languages don't have good support for Struct and we should support different formats anyway


fundthmcalculus:
TL-DR: I approve.
I fixed the `Struct()` issue in Python, but I agree with KISS principle. JSON also makes the process of consumption much easier for other languages we don't support, should somebody decide to write a wrapper library in another language. Now you're using simple FFI with primitive types. The complex types are serialized using the best method in each language.

tmarkovski:
I think the first time we'll encounter this again will be with ObjectiveC and potentially Java. ObjC needs a byte array to work with, so better to get this out the door now.
Our API on the SDK side already uses language native format (`JObject` in .NET, JS objects in JS, etc), so this change only adds cleanliness to our API

MichaelEdwardBlack (approved):

PR #69, opened 2021-07-02
Build script support

fundthmcalculus:


michaeldboyd (approved):

devops/BuildPython.ps1
```diff
@@ -1,7 +1,7 @@
 param
 (
     [string]$GitTag,
-	[switch]$UseDevelopOkapi = $false,
+	[switch]$UseDevelopOkapi = $false
```
tmarkovski: I can still see two extra spaces in there. How does that not bother you! :)

tmarkovski (approved):

devops/BuildPython.ps1
fundthmcalculus: that was, god-forbid! a tab...

PR #72, opened 2021-07-04
Reorganize and edit docs

michaeldboyd:
- added docs sample and script
- added consolidated introduction
- added updated walkthrough guide
- cleaned up the walkthroughs
- fixed broken links
- consolidated all nav + toc into sidebar


cli/sample/runthrough.sh
```diff
@@ -0,0 +1,26 @@
+
```
tmarkovski, line 2: Script is missing sha-bang

tmarkovski: The script file has no execute flag. For me, it opened in an editor. `chmod +x runthrough.sh` will fix it

cli/sample/runthrough.sh
```diff
@@ -0,0 +1,26 @@
...
+# Provider
+trinsic wallet create --description "Clinic's Wallet" --name clinic && \
+trinsic wallet create --description "Airline's wallet" --name airline;
+
+# Allison
+trinsic wallet create --name allison --security-code <code>
+
+# Issuer
+trinsic --profile clinic issuer issue --document ../vaccination-certificate-unsigned.jsonld --out ./vaccination-certificate-signed.jsonld
```
tmarkovski: May be an extra dot here

tmarkovski (commented):
I like the top level reorganization, it's an improvement.

Not sure how I feel about the individual page links showing up on the left side, instead the right. To me it suggests they are individual links to different pages, rather than section anchors. It's a little confusing. What's the reasoning behind this approach?

Cross checking with some doc sites. Microsoft uses links on left, page anchors on right
image

Similarly, Google
image

AWS
image

michaeldboyd:
I agree that it's a little confusing at first to have the table of contents added to the left hand page directory. However, I think once the user understands that's the convention, it makes future navigation a bit more efficient when looking at what a page has to offer while exploring. That's at least how it felt to me when I was trying out both approaches. 
Twilio does this that way:
image

However, after looking at more docs sites (stripe, segment) seems that they all follow the toc-on-right-side convention. 

I don't mind leaving it separate for now. It's just a feature flag if we want to turn it on again in the future

tmarkovski (approved):

PR #73, opened 2021-07-06
SDK-57 Get test from environment variable

fundthmcalculus:


michaeldboyd (approved):
LGTM

fundthmcalculus:
@michaeldboyd @tmarkovski do we have the server address in the environment variable yet? Or may I set it up with the staging env that I was using hardcoded?

tmarkovski:
> @michaeldboyd @tmarkovski do we have the server address in the environment variable yet? Or may I set it up with the staging env that I was using hardcoded?

Yes, please use staging endpoint. I don't think it's set yet.

PR #109, opened 2021-09-08
Trust Registry CLI commands

tmarkovski:
CLI commands for Trust Registry service

### DIFF
--- a/cli/src/cli.yaml
+++ b/cli/src/cli.yaml
@@ -352,6 +352,18 @@ subcommands:
         - list_credential_templates:
             about: "List credential templates"
             version: "0.1"
+  - trust-registry:
+      about: Trust Registry
+      subcommands:
+        - search:
+        - check-issuer:
+        - check-verifier:
+        - download:
+        - register-egf:
+        - register-issuer:
+        - register-verifier:
+        - unregister-issuer:
+        - unregister-verifier:
 args:
   - profile:
       long: profile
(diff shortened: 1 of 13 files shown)


fundthmcalculus (approved):

cli/src/cli.yaml
```diff
@@ -352,6 +352,125 @@ subcommands:
...
+        - search:
+            about: |
+              Search the registry for authoritative issuer and verifiers
+              using a custom query in a SQL format
+            args:
+              - query:
+                  long: query
+                  value_name: SQL query
+                  help: |
+                    Input SQL query to use to search the registry.
```
michaeldboyd: probably meant "input sql query to search the registry"

cli/src/cli.yaml
```diff
@@ -352,6 +352,125 @@ subcommands:
...
+                    Default value is "SELECT * FROM c"
+                  takes_value: true
+                  required: true
+        - check-issuer:
+            about: |
+              Check the status of an issuer for a credential type within
+              a given governance framework.
+              Returns all historical data for the given input parameters
+            args:
+              - did-uri:
```
michaeldboyd: Why shouldn't we just call this `did` ? a `did-uri` seems like it's supposed to be a URI of the DID

michaeldboyd (approved):
Just a naming comment and a typo comment

cli/src/cli.yaml
tmarkovski: Updated

cli/src/cli.yaml
tmarkovski: Updated. Feel free to make PR with any changes related to naming, text, spelling, etc. Merging this.

PR #112, opened 2021-09-09
Swift services

fundthmcalculus:


tmarkovski:
Are files under `.swiftpm` supposed to go in source control?

tmarkovski:
@drruruu Tagging you for visibility into failing PR checks due to what seems like a non-existent environment

fundthmcalculus:
@tmarkovski I'm not sure, the community consensus is mixed, since xcode likes to put things there that normally would be committed. Given Xcode's proclivity for breaking due to missing files, I figured better safe than sorry. If we keep having meaningless changes to those files, I will specifically ignore them.

But to answer your original question, `.swiftpm` original intent did *not* need to be added to git.

tmarkovski (approved):

Issue #95, opened 2021-08-06
Make sure all service SDKs parse any valid URL correctly & strip http:// etc.


fundthmcalculus:
Currently only Python is smart about this. We can extend to other languages easily
- [x] Go
- [x] Java
- [x] C#
- [x] Swift
- [x] Python
- [x] Objective C
- [x] Ruby

PR #120, opened 2021-09-15
Sdk 95 parse urls correctly

fundthmcalculus:
https://github.com/trinsic-id/server/issues/219
https://github.com/trinsic-id/sdk/issues/95


dotnet/Source/Trinsic.Tests/Tests.cs
```diff
@@ -17,6 +18,29 @@ public Tests(ITestOutputHelper testOutputHelper)
         {
             _testOutputHelper = testOutputHelper;
         }
+        
+        private static string __FILE__([CallerFilePath] string fileName = "") => fileName;
+
+        private string TestDataPath =>
+            Path.GetFullPath(Path.Join(__FILE__(), "..", "..", "..", "..", "devops", "testdata"));
```
tmarkovski, line 25: Yikes. Is there a better of passing the path? Maybe from ENV?

fundthmcalculus: Possibly, but then we require environment variables to be known and set for testing. I'd rather compile it in so it's known. Also this eliminates `cwd` problems.

tmarkovski (approved):

dotnet/Source/Trinsic.Tests/Tests.cs
tmarkovski: We already require them for AppConfig Conn String. Can you instead add the files in the csproj with the test data raference?

```
<Include Path="../../../../..../testdata/file.jsonld" CopyToOutput=true />
```

or whatever the correct syntax is?

fundthmcalculus: I can do that for C#, most other languages don't have that kind of build system. :/

PR #129, opened 2021-09-21
trustregistry service for ts web package

tmarkovski:


web/src/TrustRegistryService.ts
```diff
@@ -0,0 +1,118 @@
...
+  RegisterIssuerRequest,
+  RegisterIssuerResponse,
+  RegisterVerifierRequest,
+  RegisterVerifierResponse,
+  RemoveFrameworkRequest,
+  RemoveFrameworkResponse,
+  SearchRegistryRequest,
+  SearchRegistryResponse,
+  UnregisterMemberRequest,
+  UnregisterMemberResponse,
```
fundthmcalculus, line 19: is there an `import { * } from ` clause that would be cleaner?

fundthmcalculus (approved):

web/src/TrustRegistryService.ts
tmarkovski: Nope, that would be convenient. There is `import * as proto` but that would require prefixing each type with `proto.`

PR #138, opened 2021-09-23
move services to interface

fundthmcalculus:


sethjback (approved):

go/services/services.go
```diff
@@ -55,46 +58,60 @@ func (s *ServiceBase) SetProfile(profile *sdk.WalletProfile) error {
...
+	if err != nil {
+		return err
+	}
 
 	invokerKey := okapiProto.JsonWebKey{}
 	err = proto.Unmarshal(profile.InvokerJwk, &invokerKey)
-	if err != nil { return err }
+	if err != nil {
+		return err
+	}
```
fundthmcalculus, line 69: I don't like this format, but if this is the `golint` and `gofmt` approved model, we'll use it.

go/services/services.go
```diff
@@ -24,19 +25,21 @@ type ServiceBase struct {
...
 
-type ServiceBaser interface {
-	GetContext() context.Context
-	GetMetadata() metadata.MD
-	SetProfile(profile sdk.WalletProfile)
+type Service interface {
+	GetContext() (context.Context, error)
+	GetMetadata() (metadata.MD, error)
+	SetProfile(profile *sdk.WalletProfile) error
 }
```
fundthmcalculus, line 32: Needs to be `ServiceBase` for consistency with other languages.

fundthmcalculus: Or is the struct `ServiceBase`?

PR #141, opened 2021-10-06
credential template node sdk with sample

MichaelEdwardBlack:
Credential templates added to the node sdk with a sample that goes through creating and searching for a template and then issuing a credential based on that template


tmarkovski (approved):

fundthmcalculus:
ignore those build errors, those are conflation of things that aren't relevant.

PR #148, opened 2021-10-21
Refactor protos to add versioning and buf.build integration

tmarkovski:


fundthmcalculus (approved):
I will follow up on the other languages shortly.

PR #156, opened 2021-10-26
Remove test for default port, as it's required by gRPC channel

### LINKED ISSUE #157: Parse url with IP Addresses correctly
**Describe the bug**
https://github.com/trinsic-id/sdk/pull/156
Fails on `http://20.75.134.127:80`
We should allow parsing of IP addresses, not error out. We should still warn if we don't provide an explicit port.

tmarkovski:


fundthmcalculus (approved):
https://github.com/trinsic-id/sdk/issues/157

PR #199, opened 2021-11-12
Java tls prod endpoints

fundthmcalculus:


dotnet/Trinsic.Tests/Tests.cs
```diff
@@ -0,0 +1,118 @@
+using System;
```
tmarkovski, line 1: I think this entire file got merged incorrectly. It no longer exist in head, folder got renamed

tmarkovski (approved):

PR #209, opened 2021-11-17
add code snippet annotations for dotnet

lucasamonrc:


lucasamonrc:
I have added the snipped annotations in [/dotnet/Tests/Test.cs](https://github.com/trinsic-id/sdk/blob/main/dotnet/Tests/Tests.cs) as requested; however, I don't think that the snippets that were included in [/docs/dotnet/vaccination-net.md](https://github.com/trinsic-id/sdk/blob/main/docs/dotnet/vaccination-net.md) are representative of the steps described in the docs. I tried to change the `Test.cs` in an attempt to make them more representative, but it caused the tests to fail, so I reverted those changes.

docs/dotnet/vaccination-net.md
```diff
@@ -33,24 +33,24 @@ dotnet add package Okapi.Net --prerelease
 
 ## Configure services
 
-Create a reference to the wallet service that points to your ecosystem service. You should have received this URL with your ecosystem setup. In your `Program.cs` file add the following line replacing the placeholder with your URL:
+Create a reference to the wallet service that points to your ecosystem service. You should have received this URL with your ecosystem setup. In your `Program.cs` file add the following line replacing the `null` argument with your URL (ignore the `serverConfig` argument):
```
tmarkovski, line 36: This is not exactly accurate anymore. We won't be giving out users URL, they would just the default that comes with the SDK, which is `prod.trinsic.cloud`. We will most likely need to update this anyway, once the AccountService gets done, in which case the init code would look a little different.

fundthmcalculus: Default cloud config is in each language SDK already.

fundthmcalculus (approved):

PR #212, opened 2021-11-18
Support for building Nix package bundle 

tmarkovski:
Adds a vendored bundle of the CLI folder for use with Nix packages


.github/workflows/release-rust.yml
```diff
@@ -99,3 +99,54 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Generate Package Version
+        run: |
+          $json = Invoke-WebRequest 'https://api.github.com/repos/trinsic-id/sdk/releases/latest' `
+            -Headers @{ "Authorization" = "Token $env:API_GITHUB_TOKEN"} | ConvertFrom-Json
+          $version = if ([string]::IsNullOrEmpty($env:MANUAL_VERSION)) { $json.tag_name.Trim("v") } else { $env:MANUAL_VERSION.Trim("v") }
+          echo "RELEASE_VERSION=$($json.tag_name)" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf-8 -Append
+          echo "PACKAGE_VERSION=$version" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf-8 -Append
```
fundthmcalculus: this powershell script keeps getting copy-pasted, which to me is a code smell. Is there a way to make it a shell script that we call instead?

tmarkovski: It can be made into a custom GitHub action. It's just faster to copy/paste it :(

tmarkovski: There, it's now a custom action

fundthmcalculus (approved):

PR #215, opened 2021-11-19
Update dotnet SDK with account service

tmarkovski:


tmarkovski:
CI won't pass until we deploy server to staging

.github/workflows/build-dotnet.yml
```diff
@@ -20,10 +20,7 @@ on:
 jobs:
   test:
     name: Test
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+    runs-on: macos-latest
```
fundthmcalculus, line 23: We should test on all platforms.

fundthmcalculus (approved):
See comment

.github/workflows/build-dotnet.yml
tmarkovski: We should, but the other platforms don't support `net6.0-ios` target, so we need to separate that out first.

PR #239, opened 2021-11-30
Sdk 235 golang accountservice

fundthmcalculus:


sethjback (approved):

go/services/security.go
```diff
@@ -0,0 +1,53 @@
...
+
+	proof, err := okapi.Oberon().CreateProof(&okapiproto.CreateOberonProofRequest{
+		Data:  profile.AuthData,
+		Token: profile.AuthToken,
+		Nonce: nonceBytes,
+	})
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("Oberon proof=%s,data=%s,nonce=%s",
```
tmarkovski: Can you please add a static `ver=1` parameter as well?

go/services/storage.go
```diff
@@ -0,0 +1,30 @@
...
+
+type ProfileProvider interface {
+	Get(name string) (sdk.AccountProfile, error)
+	Save(name string, profile sdk.AccountProfile) error
+}
+
+type FileProfileProvider struct {
+}
+
+type KeychainProfileProvider struct {
```
tmarkovski, line 13: FYI, these are not required. `KeychainProfileProvider` is intended for use with mobile platforms. Since Go isn't a mobile platform, it should only implement `FileProfileProvider`. These are not designed yet, so we're just letting users handle the profiles themselves.

go/services/security.go
fundthmcalculus: Woops

go/services/storage.go
fundthmcalculus: I figured something like that was the case, so I decided to stub it out as future work.

tmarkovski (approved):

sethjback (approved):

PR #241, opened 2021-12-01
Ruby account service

fundthmcalculus:
Brings Ruby up to date with latest okapi, accountservice, etc


ruby/lib/security.rb
```diff
@@ -0,0 +1,39 @@
...
+      Base64.urlsafe_encode64(Google::Protobuf::encode(message))
+    end
+
+    def get_auth_header(account_profile, message)
+      if account_profile.protection.enabled
+        raise "The token must be unprotected before use"
+      end
+
+      request_bytes = Google::Protobuf::encode(message)
+      request_hash = Digest::BLAKE3.digest(request_bytes)
```
tmarkovski: When the request is serialized to an empty object, this function fails, as the hash input expects non-empty buffer. In other languages, we just don't hash if `request_bytes` is empty, and set the empty array directly in the `nonce`.
Just want to make sure this is the case here, as well.

ruby/lib/trinsic_services.rb
```diff
@@ -0,0 +1,241 @@
...
+  Account_V1 = Services::Account::V1
+  Credentials_V1 = Services::Verifiablecredentials::V1
+  Provider_V1 = Services::Provider::V1
+  TrustRegistry_V1 = Services::Trustregistry::V1
+  Wallet_V1 = Services::Universalwallet::V1
+
+  def self.trinsic_test_server
+    server_endpoint = ENV["TEST_SERVER_ENDPOINT"]
+    server_port = ENV["TEST_SERVER_PORT"] || "443"
+    server_usetls = ENV["TEST_SERVER_USETLS"] || "true"
```
tmarkovski: The env name we usually use here is `TEST_SERVER_USE_TLS`

ruby/test/trinsic_service_test.rb
```diff
@@ -2,69 +2,41 @@
...
-    provider_service = Trinsic::ProviderService.new(server_address)
+    account_service = Trinsic::AccountService.new(nil, Trinsic::trinsic_test_server)
+    account_profile = account_service.sign_in(nil).profile
+    provider_service = Trinsic::ProviderService.new(account_profile, Trinsic::trinsic_test_server)
 
     invite_request = Services::Provider::V1::InviteRequest.new(:description=>"I dunno",
                                                           :email=>"scott.phillips@trinsic.id")
-    invite_response = provider_service.await.invite_participant(invite_request)
-    assert(invite_response != nil)
+    # invite_response = provider_service.invite_participant(invite_request)
```
tmarkovski, line 35: You will be spammed with token codes, you should use invalid email address

ruby/lib/trinsic_services.rb
fundthmcalculus: Good catch!

tmarkovski (approved):

PR #260, opened 2021-12-06
[dotnet] Standardized ctors. Sync api.

tmarkovski:


dotnet/Tests/Tests.cs
```diff
@@ -64,7 +64,7 @@ public async Task TestWalletService()
         // Read the JSON credential data
         var credentialJson = await File.ReadAllTextAsync(VaccinationCertificateUnsigned);
         // Sign the credential using BBS+ signature scheme
-        var credential = await credentialsService.IssueCredential(document: JObject.Parse(credentialJson));
+        var credential = await credentialsService.IssueCredentialAsync(document: JObject.Parse(credentialJson));
```
fundthmcalculus, line 67: Are we going to have a non-async api? If not, I don't think we need `Async` on the end of every function

fundthmcalculus (approved):

dotnet/Tests/Tests.cs
tmarkovski: We are. This PR introduces it. This is C# specific, we don't need to support sync and async api in all frameworks. C# allows that by providing sync and async through the Grpc tooling.

PR #283, opened 2021-12-15
Update node and web with strict TS, service options

tmarkovski:
- added support for older ES in web sdk
- updated web and node TS config to use strict type checking
- service ctor uses options object, instead individual args
- samples moved in examples repo


fundthmcalculus (approved):
We should make `ServiceOptions` a protobuf based object, so that we can keep it consistent across all languages.

tmarkovski:
@MichaelEdwardBlack I'm not sure if you want to test before we merge if it works with your wallet app, but would be nice to get some kind of feedback that this helped the problem, before we publish new package version.

MichaelEdwardBlack (approved):
Tested with the react app and seems to work great! Thank you!

PR #287, opened 2021-12-21
Credential templates and provider proto update

tmarkovski:


fundthmcalculus (approved):
I will make the other language namespace header changes.

fundthmcalculus (approved):

PR #299, opened 2021-12-24
Credential template tests. Service creation refactor

tmarkovski:


fundthmcalculus (approved):
Will do the same for the other languages

PR #309, opened 2021-12-27
Golang release tag action

### LINKED ISSUE #282: publish go release tag as a release action
As reported in https://github.com/trinsic-id/sdk/issues/281, we need to have an automatic tag in the go release action so that people can `go get` by version instead of commit id.

fundthmcalculus:


.github/workflows/build-golang.yml
```diff
@@ -1,4 +1,4 @@
-name: Go
```
tmarkovski, line 1: Why Golang? https://go.dev/doc/faq#go_or_golang

.github/workflows/release-golang.yaml
```diff
@@ -0,0 +1,22 @@
...
+      packageVersion:
+        description: 'Version to build'
+        required: true
+        default: 'v1.0.0'
+  release:
+    types: [ published ]
+
+jobs:
+  publish_go_tag:
+    name: Tag golnag
```
tmarkovski: Typo

fundthmcalculus: LOL. Fixed.

.github/workflows/build-golang.yml
fundthmcalculus: Specifically because "go" in English is a verb, and I find it confusing and slightly annoying. "go bindings" would be fine too, but that is inconsistent as well.

tmarkovski: I agree, especially when googling "go", though the use of the Golang as official name is incorrect and a little odd.

tmarkovski (approved):

PR #313, opened 2021-12-29
make Java build with the proper location

fundthmcalculus:
- [x] Blocked by this: https://github.com/trinsic-id/okapi/pull/320
- [x] Then update Java jar wrapper
- [x] Then update `build.gradle` here


fundthmcalculus:
This has a failing test due to invitation status not having ecosystem id set. That will be a separate PR. I want to get this in first so that default build actually works.

tmarkovski (approved):

PR #386, opened 2022-01-19
Ts/vaccine demo

### LINKED ISSUE #384: Implement VaccineDemo for Typescript
@lucasamonrc please create a typescript/javascript version of the vaccine credential demo. Follow the format in C# or Python, either is up to date.

This will just be a really long unit test, or however makes the most sense to you. The end requirements are:
1. It has to run when the unit tests run.
2. It's easy to follow to code inject for the documentation.
3. Needs to be in both `node` and `web`

lucasamonrc:


fundthmcalculus (approved):
thanks for writing the tests and docs! We will have to merge the docs into `mdb/learn`, but the tests are great where they are.
@michaeldboyd shall we integrate into your branch?

PR #387, opened 2022-01-20
Make all node tests work

fundthmcalculus:
Refactor node path to use common sample data
Refactor to use string handling


web/src/ServiceBase.ts
```diff
@@ -31,13 +31,13 @@ export default abstract class ServiceBase {
...
     }
 
-    var requestData = request.serializeBinary();
-    var requestHash = new ArrayBuffer(0);
+    const requestData = request.serializeBinary();
+    let requestHash = new ArrayBuffer(0);
 
     if (requestData.length > 0) {
-      requestHash = await crypto.subtle.digest("SHA-256", request.serializeBinary());
+      requestHash = await crypto.subtle.digest("sha256", request.serializeBinary());
```
fundthmcalculus:
```suggestion
      requestHash = await crypto.subtle.digest("SHA-256", request.serializeBinary());
```

tmarkovski (approved):

PR #400, opened 2022-01-24
go doc updates with some dotnet doc updates as well.

fundthmcalculus:


docs/go/index.md
```diff
@@ -2,12 +2,20 @@
...
-TODO
-
-## Configuration
+## Installation and Configuration
+* NOTE: Windows installation is possible, but requires jumping through a few hoops. More details to come later.
+* Make sure you have the okapi native binaries installed on your machine. Default installation instructions are in the [readme.md here](https://github.com/trinsic-id/okapi/)
+* You need to have [CGO configured](https://pkg.go.dev/cmd/cgo) for the okapi native binary link
+* Issue the following command:
+```bash
+go install https://github.com/trinsic-id/sdk/go
```
sethjback: the `go get` and `go install` commands all leave off the URL scheme (`https://` )

devops/build_sdks.py
```diff
@@ -139,6 +148,25 @@ def build_dotnet_docs(args) -> None:
                      cwd=current_dir).wait()
 
 
+def build_go_docs(args):
+    # https://github.com/posener/goreadme
+    # go get github.com/posener/goreadme/cmd/goreadme
+    goreadme_args = r'-recursive -functions -methods -types -variabless'  # Yes, that's a duplicated s, it's on purpose.
+    doc_path = abspath(join(get_language_dir('docs'), 'reference', 'go'))
```
sethjback, line 155: I think this is great to have a readme in the repo. - the next step would be to comment the code so that public [go docs](https://pkg.go.dev/search?q=trinsic) and the generated readme have more detailed information.

This can be done later - I've always found coding with the volume of comments required to make pretty documentation to be a pain :-), but it looks good and documentation at https://pkg.go.dev will be one of the first places golang dev will look.

sethjback (changes requested):

devops/build_sdks.py
sethjback: This is a comment only - no changes requested here

docs/go/index.md
fundthmcalculus: Thanks!

devops/build_sdks.py
fundthmcalculus: Agreed on all counts. My goal is to get the generated docs procedure in place, and then flesh out the content.

docs/go/index.md
fundthmcalculus:
```suggestion
go install github.com/trinsic-id/sdk/go
```

sethjback (approved):

PR #406, opened 2022-01-26
Updated android and java install instructions

fundthmcalculus:


docs/android/index.md
```diff
@@ -1,15 +1,23 @@
...
 
 ## Installation
-1. Install the Trinsic SDK package from the [Github Packages](https://github.com/trinsic-id/sdk/packages/940563)
+1. Download the trinsic SDK package (`services-X.X.X.jar`) from the [Github Packages](https://github.com/trinsic-id/sdk/packages/940563).
+2. Download the trinsic okapi package (`okapi-X.X.X.jar`) from the [Github Packages](https://github.com/trinsic-id/okapi/packages/919613).
+3. Put the jar files into the appropriate folder (`./android/app/libs`) - this allows gradle to package them in, since public anonymous gradle access to Github Maven Packages is not supported.
+4. Download the latest [release artifact](https://github.com/trinsic-id/okapi/releases/latest) asset called `libs.zip`
+5. Extract it, under `libs/android` you should find the various android device (and emulator) architectures supported.
+6. Copy the `libokapi.so` from the corresponding directory to the corresponding architecture directory under `android/app/src/main/jniLibs`.
```
michaeldboyd: Maybe we should leave commands on this? just so its explicit.

michaeldboyd (approved):

docs/android/index.md
fundthmcalculus: I think that people can see the folders exactly match, and surmise from there. Also, what if (when) we add another android architecture support?

PR #407, opened 2022-01-26
go and typescript walkthrough support

fundthmcalculus:


michaeldboyd (approved):
Not a go expert, but the codeincludes LGTM

PR #415, opened 2022-01-27
Assign milestone to PR

tmarkovski:


.github/milestone.yml
```diff
@@ -0,0 +1,6 @@
+base-branch:
+  - "(main)"
+  - "2022 Q1"
+
+head-branch:
+  - "2022 Q1"
```
fundthmcalculus: will `2022Q1` have to be updated every quarter?

fundthmcalculus (approved):

.github/milestone.yml
tmarkovski: Yeah, I can't find an action that automatically parses the milestone, checks for date ranges, ensures the date range is in current date, and if multiple milestones are found, decide which one to attach.

PR #419, opened 2022-01-27
Build proto files on github actions

fundthmcalculus:


tmarkovski:
Can you post what is the intended behavior? If there are changes to the proto files, this will automatically create a PR?

tmarkovski (approved):

fundthmcalculus:
It will auto create a PR. That doesn't work off existing PRs, but thats a limitation of git

PR #424, opened 2022-02-01
SDK 345 - Expose proper public constructors: python, go, C#, Java

fundthmcalculus:


dotnet/Trinsic/AccountService.cs
```diff
@@ -41,6 +31,16 @@ public AccountService(AccountProfile accountProfile, GrpcChannel channel)
         : base(accountProfile, channel) {
         Client = new(Channel);
     }
+    
+    internal AccountService(ServerConfig serverConfig)
```
tmarkovski, line 35: What's the reasoning for hiding this ctor?

fundthmcalculus: Unit tests, hence `internal`.

tmarkovski: This prevents end users from ever passing server endpoints, even if they have their own instance or we ask them to test against staging for whatever reason. Not sure limiting this is a good approach.

fundthmcalculus: Right, @sethjback and I had a discussion surrounding this last night, that we want to ensure people can hit staging if we need them to do so, but never do so accidentally. Seth and I think the best manner for that is to set the environment variables `TRINSIC_SERVER_ADDRESS`, `TRINSIC_SERVER_PORT`, and `TRINSIC_SERVER_USE_TLS` and let them automatically override. I'm trying to reduce the chance of a copy-paste mistake from the samples.

fundthmcalculus: That being said, if this is a solution to a negligible problem, I'm happy to discard it.

tmarkovski: Samples should be coded as if they're running against production in a way that they can copy/pastable.
The ENV variable setting can still be applied where it makes sense, but thinking more broadly, the SDK runs in platform where there aren't environment variables, like mobile or browser. This means the solution should be more generalized, imo.

fundthmcalculus: Agreed on all counts. Let me close this PR for now, until we have agreement on how to handle it.

PR #427, opened 2022-02-02
Improve build process - run all actions when `devops` changes, only download required binaries

fundthmcalculus:
This fixes issues with java not building, actions timing out due to artifact download size (ios and android variants are the biggest offenders), not running required actions when devops python scripts change, python script now allows language-specific selection for isolation.


fundthmcalculus:
Ruby on macos has issues due to environment variable permissions. I will update the ruby okapi package in a separate PR to address this.

tmarkovski (approved):

PR #433, opened 2022-02-07
couple minor updates for M1 support

fundthmcalculus:


ruby/Gemfile
```diff
@@ -1,2 +1,2 @@
-source "https://rubygems.org"
-gemspec
\ No newline at end of file
+source 'https://rubygems.org'
```
allibell, line 1: paste (2)

python/tests/test_trinsic_services.py
```diff
@@ -15,6 +16,9 @@
 # Due to some issues with python and async io test cases, we have to run each sample in a separate asyncio event loop.
 
 class TestServices(unittest.TestCase):
+    def test_python_platform(self):
+        print(f"Running on: {platform.platform()}")
```
allibell, line 20: should we have any asserts here?  like non-null or something?  or is this just for debugging utility

allibell (commented):
genuinely curious about how double quotes might relate to M1 Ruby things, otherwise LGTM!

ruby/Gemfile
fundthmcalculus: Rubocop style guidelines. Single quotes in Ruby make it a string literal, double quotes allow for string interpolation and stuff like that. This makes it clear it's just a string of characters, nothing exciting.

python/tests/test_trinsic_services.py
fundthmcalculus: No, this is a debug test (that I plan to replicate everywhere) that will enable us to ensure what platform the code *thinks* it's running on, eg for M1 (native) vs M1 (emulated intel).
If it makes you feel better, I can `assert platform.platform() is not None`. 😁

allibell: Nah - just wanted to check you didn’t intend to have legit unit test stuff here and then lose some logic in the shuffle.  That makes sense!

allibell (approved):

PR #437, opened 2022-02-10
Create bug-report template

fundthmcalculus:
Eventually this should have a better formatting, but I want a sample template for our users to see when they go to create an issue.


tmarkovski (approved):

tmarkovski:
Something's wrong, it's not showing up.

PR #453, opened 2022-02-27
[dotnet] Endpoint and proto definitions update (breaking)

tmarkovski:
- Removed the use `JsonPayload` message with `oneof` implementation. Only uses `string` as JSON
- Added `CreateProofDirect` endpoint to VC service for proof creation directly from a request


fundthmcalculus:
Does this require work in all other languages as well? If so, let's create an SDK change request ticket.

tmarkovski:
> Does this require work in all other languages as well? If so, let's create an SDK change request ticket.

We will, once we've identified all changes we'd like in this push. There's few issues still under discussion that we need to resolve first. This PR is not to be merged, just here for visibility and comments.

fundthmcalculus (approved):
Do you want me to make corresponding PRs for the other languages against this PR, or separately?

dotnet/Tests/Tests.cs
```diff
@@ -234,11 +234,16 @@ public class Tests
...
+        var accountService = new AccountService(_serverConfig);
+        var profile = await accountService.SignInAsync();
+        var providerService = new ProviderService(profile, accountService.Channel);
+        
+        var invitationResponse = await providerService.InviteParticipantAsync(new());
+
+        invitationResponse.Should().NotBeNull();
+        invitationResponse.InvitationCode.Should().NotBeEmpty();
+        
+        await Assert.ThrowsAsync<Exception>(async () => await providerService.InvitationStatusAsync(new InvitationStatusRequest()));
```
fundthmcalculus, line 246: At the risk of being pedantic, we're mixing assertion forms. I prefer `Assert.NotNull()` myself, but that's just me.

tmarkovski:
> Do you want me to make corresponding PRs for the other languages against this PR, or separately?

Yes, after we consider the naming changes and API changes final. I wouldn't say they're final yet. The profile management part is largely undecided yet.

fundthmcalculus:
@tmarkovski shall I commence the other languages coming into parity?

tmarkovski:
Sure, but also let's finalize first which additional updates we want to make. This PR doesn't assume final changes. Comments at https://github.com/trinsic-id/sdk/issues/460

fundthmcalculus:
> Sure, but also let's finalize first which additional updates we want to make. This PR doesn't assume final changes. Comments at #460

Commented there.

PR #457, opened 2022-03-01
Use an options structure for construction

fundthmcalculus:
This is to eliminate the side effect of ignoring the `ServerConfig` if the channel is provided. It will throw an error if both the `ServerConfig` and `Channel` are provided as arguments.

Go: does not allow function overloading, so an `Option` structure is used.
Python: Use keyword arguments only, so user has to be explicit on what they set.


go/services/services.go
```diff
@@ -16,6 +16,12 @@ import (
 
 type Document map[string]interface{}
 
+type ServiceOptions struct {
+	profile *sdk.AccountProfile
+	config  *sdk.ServerConfig
+	channel *grpc.ClientConn
+}
```
sethjback, line 23: Config structs work with two caveats:

1. Can we make sure, to the extent possible, there are reasonable defaults? I.e., if you don't pass in anything to a function, it should just work, and to the extent it can't, then that option should be required to be provided.

2. I like the option function pattern, as it lets you craft the "factory" functions to embody the above assumption about reasonable defaults:

```go

func NewProviderService(opts []Options) (*ProviderService, error)

```

If no options are provided, use the defaults, otherwise only override the provided ones.

To the extent that there are mutually exclusive options you can just check after applying all the functions whether both are set and return an error.

fundthmcalculus: Good call.
1. We currently do that. Defaults are no profile (will error that profile must be set) and create new connection to trinsic production server.
2. Good point. I experimented with this, and it ended up being much more verbose than `ServiceOptions {profile: profile, channel: accountService.channel}` or `ServiceOptions{}` if you are instantiating `AccountService` with no profile and production.
3. We do check for providing both `config` and `channel`. It will throw an error as being ambiguous.

sethjback: Yeah, if you are setting all the options it can get pretty verbose, but then I've seen things like `withServiceOptions()` or even just calling the `NewProviderService` and then replacing the options directly.

sethjback (approved):

PR #467, opened 2022-03-08
Sdk 460 update python, and generated proto files

fundthmcalculus:


dotnet/Trinsic/ProviderService.cs
```diff
@@ -81,7 +81,7 @@ public ProviderService(ServiceOptions options)
         if (string.IsNullOrWhiteSpace(request.Name)) throw new("Field 'name' must be specified");
         request.Details ??= new();
         
-        return Client.CreateEcosystem(request);
```
tmarkovski: CreateEcosystem is a non-authenticated endpoint, shouldn't pass metadata.

fundthmcalculus: What about `CreateEcosystemAsync` then? Should that also not be authenticated? I'll need to update the other languages with any other unauthenticated endpoints.

tmarkovski: Yes. `Account.SignIn` and `Provider.CreateEcosystem` (with their async variants) are the two endpoints that have no authentication.

tmarkovski (approved):

dotnet/Trinsic/ProviderService.cs
sethjback: So, I might suggest that we don't keep a list of authenticated endpoints in the clients - that seems like it could get to be a burden to update across all languages and calls going forward. 

What about just checking for the auth token and if present adding an "authorization" header. The `server-api` has all the calls that require auth and will quickly return a 401 if the header is not present.

fundthmcalculus: Good call @sethjback , but I do exactly the opposite. I provide authentication on every call UNLESS it's something predetermined to not be authenticated. The problem is during signin, the metadata builder needs to know if it should not throw the "account profile not set" error.

sethjback: @fundthmcalculus right, so I guess what I was suggesting was not worrying about throwing a "profile not set" error in the clients at all - basically, add the auth header if its there, if not make the call without it, and let the `server-api` return a `401` error. Seems like the net result is the same - the developer/sdk user is informed they need to set an auth token to make the call, but we won't have to maintain a list of authorized calls in two places.

The only reason I could see refusing to make the call at the client level would be to save the round trip, but the added complexity of maintaining an authorized endpoint list in every sdk language implementation doesn't seem to justify whatever benefits we get from that.

Thoughts?

fundthmcalculus: I get what you're saying Seth. I'd rather save the round trip and let the developer know immediately what's going on and why. We aren't really keeping a list in 2 places, we're adding it on every call no matter what unless we explicitly say not to.
TM is talking about having a bunch of unauthenticated endpoints I would agree that it's probably something we don't want to handle at the client level. However, if it's only these 2 the create ecosystem and sign in, I'm fine managing it client side.

tmarkovski: Yeah, I think it's only these 2 endpoints, everything else should be authenticated. We already have the code setup to throw on client side if token is not set. This helps eliminate setup issues, but also ensures that the SDK can throw reliably even if there's no connectivity. This error should indicate to the developer that they haven't setup the options correctly.

sethjback: Ok, that's fine - I just see 2 endpoints times however many languages we support, so if we add or change 1 endpoint in the server to update authentication it is a change in a significant number of places in this code base.

I'm glad I asked though because I just removed that from the golang sdk - I'll add it back :smile:

fundthmcalculus: @sethjback before you go removing things in the SDKs, please ask me first. XD

sethjback: >  before you go removing things in the SDKs, please ask me first. XD

Yep! And for sure any changes in my PRs are just suggestions and always open to discussion / push back!

Issue #460, opened 2022-03-04
Proposed changes to SDK


tmarkovski:
# Objective

With the changes to `AccountProfile` and it's use in server configurations, we have an opportunity to consolidate the service options and lessen the complexity of service construction.

# Proposal

Here are some changes I propose we do.

## Changes to Account Profile 

- [x] Update the `SignIn` and `CreateAccount` endpoints to return a string that's a base64url encoded string of the serialized byte array of the token.
- [x] Change the reference of `AccountProfile` or `Profile` to `AuthToken` or `AuthenticationToken`. This makes it clearer as to what this profile represents and how it relates to commonly used API tokens, access tokens, etc.
- [x] When `SignIn` or `CreateAccount` is called, if the response sends an unprotected profile, set the token for the current service instance as active. This won't happen if the user is using an email to sign in, which should be most of the cases, so they will need to set it again. I really don't like this experience, and hope we can find something that resembles a more common developer flow.
- [x] Ensure that when passing an auth token that is protected, the SDK throws an error with a message indicating how this can be unprotected, with a documentation link explaining why this protection is in place

## Changes to Service Options

- Add `AuthToken` to `ServiceOptions` (proto) so that tokens can be added in code configurations in server scenarios.
- Move `ServerConfig` to `ServiceOptions` (sdk proto). This really just flattens the `ServiceOptions` instead it having a complex structure. Example:

```proto
message ServiceOptions {
  string server_address = 1; // default = "prod.trinsic.cloud"
  int32 server_port = 2; // default = 443
  bool server_use_tls = 3; // default = true
  string auth_token = 4; // default = <empty>
  string default_ecosystem = 5; // default = "default"
  // etc
}
```
- [gRPC Channel](https://grpc.io/docs/what-is-grpc/core-concepts/#channels) reuse. I know there's a separate discussion on this, we can leave this feature intact, but I do want to point out that this can be filled with side effects and for the most part it's really unnecessary that we expose the actual gRPC channel in the services. It introduces complexity in object and service construction, there's the option that users can simply toggle options on it that would make the channel incompatible with the backend (turn of compression). It's also a big surface for security related bugs.

## Changes to accommodate ecosystem updates

- [x] Add `DefaultEcosystem` to the `ServiceOptions` so clients can configure their SDKs to point to a specific ecosystem instance
[...]

PR #473, opened 2022-03-09
Updates to the golang SDK

sethjback:
Initial pass at implementing changes discussed in #460 along with the changes to the ecosystem/wallet handling

This should not be considered complete - there are additional changes that need to be made to bring the service calls in line, but there were discussion items to be resolved before completing (see #474)

#### Update to use ServiceOptions

* Updates `proto/sdk/options/v1/options.proto` to put the `ServiceOptions` package in same package as the rest of the services
* `NewSerivceBase` now takes a `ServiceOptions` struct and uses it to configure itself
* Profile get and Set now take and return strings
* All services that use service base now take `ServiceOptions` in their constructor functions
* `services.go` implements the functional options pattern for building the `ServiceOptions` struct
* Reasonable defaults are set for all options

#### Updates to BaseService

* Get/Set profile now expect the auth token string
* `BuildMetadata` now decodes auth token into profile before passing to security service
* Uses security provider interface rather than the direct base oberon implementation
* removed experimental grpc methods to force connections at create but relies on the expected lazy connect behavior
* Consolidated grpc client creation into single method since the new `ServerOptions` has endpoint/port/tls separate (no longer necessary to parse url for this data)
* It has methods to Get and Set the grpc client connection

#### TestUpdates

* tests have been updated so that all active tests pass on dev
* these are not exhaustive. Where there was some outstanding questions the failing tests were commented out so we can go back and update

#### Code Commenting

* The `AccountService` has a first pass at the minimal comments that would need to be in place for `golint` to not complain. Currently the actual unit tests are not running since the action is failing on the `golint` step. This PR only adds comments to the `AccountService` since there is a discussion item that needs to be resolved before commenting the rest of the codebase.


go/services/account_service.go
```diff
@@ -20,78 +25,109 @@ func NewAccountService(options ServiceOptions) (AccountService, error) {
...
 }
 
+// AccountService wraps all the functions for interacting with accounts
 type AccountService interface {
 	Service
-	SignIn(userContext context.Context, details *sdk.AccountDetails) (*sdk.AccountProfile, sdk.ConfirmationMethod, error)
-	Unprotect(profile *sdk.AccountProfile, securityCode string) (*sdk.AccountProfile, error)
-	Protect(profile *sdk.AccountProfile, securityCode string) (*sdk.AccountProfile, error)
+	// SignIn returns an encoded auth token
+	SignIn(userContext context.Context, details *sdk.AccountDetails, inviteCode, ecosystemID string) (string, sdk.ConfirmationMethod, error)
```
tmarkovski, line 32: In other SDKs we're also passing the actual `SignInRequest` here as well, and just relying on devs to construct them.
I do like how this method return indicates if the token requires unblinding.
I've no strong opinion on this deconstructed choice, just FYI. One argument for using single parameter request is that we can make additions/changes to it, without breaking the exposed API.

tmarkovski (approved):
LGTM

go/services/account_service.go
sethjback: Based on our discussion #474 I think the intent is to standardize on passing the actual sdk request and response objects. I am taking a second pass based on the answers in that issue and will update all the calls to use this approach

PR #478, opened 2022-03-10
Update to Golang SDK

### LINKED ISSUE #474: Proposed Golang SDK Updates
### Purpose

The goal of this issue is to open a discussion around a couple of consistency / housekeeping items for the golang codebase.

#### Private Base Implementations

Currently ever service is defined in an interface and then a base implementation is provided. Those base implementations are not commented in the standard way for godoc to build documentation and thus tools like `golint` are complaining and ultimately failing.

It is a pattern to keep the base implementations of interfaces private (unexported) thus not requiring the same level of documentation.

The question to be resolved:

Is there a reason to have the base implementations exported? If not they can be made private and there is no further work required. If yes, we need to comment the code base appropriately.

#### Function Call Arguments

Currently we have a mix of passing in the Server API request struct directly and passing in fields that then get validated and an api request is built from them.

grpc request struct:

```golang
func (c *CredentialBase) CheckStatus(userContext context.Context, request *sdk.CheckStatusRequest)
``` 

request fields:

```golang
func (c *CredentialBase) VerifyProof(userContext context.Context, proofDocument Document)
``` 

Either way has it's pros/cons, we just need to be consistent.

#### Unit vs. Integration testing

Currently there is a mix of unit and integration tests in the test files. I would like to propose we keep the tests in the `_test` files to unit tests and things we can mock, and have a separate process in place for doing full blow integration testing against a deployed environment.

sethjback:
Closes #474

* Makes base implementations of the interfaces private
* Comments interface definitions
* Consistent use of sdk service request objects across calls
* Service constructors use `Options` struct that wraps `ServiceOptions` and the grpc client connection to allow connection sharing between clients

### DIFF
--- a/go/services/services.go
+++ b/go/services/services.go
@@ -6,37 +6,45 @@ import (
 	"strings"
 
 	sdk "github.com/trinsic-id/sdk/go/proto"
+	"google.golang.org/grpc"
 	_ "google.golang.org/protobuf/types/known/structpb"
 )
 
+//Options for configuring the sdk
+type Options struct {
+	ServiceOptions *sdk.ServiceOptions
+	Channel        *grpc.ClientConn
+}
+
 // NewServiceOptions returns a service options configuration with the provided options set
-func NewServiceOptions(opts ...Option) (*sdk.ServiceOptions, error) {
-	so := &sdk.ServiceOptions{}
+func NewServiceOptions(opts ...Option) (*Options, error) {
+	options := &Options{
+		ServiceOptions: &sdk.ServiceOptions{
+			DefaultEcosystem: "default",
+		},
+	}
 
 	// Default to production
-	WithProductionEnv()(so)
-
-	// with the default ecosystem set
-	so.DefaultEcosystem = "default"
+	WithProductionEnv()(options)
 
 	// set user defined options
 	for _, o := range opts {
-		err := o(so)
+		err := o(options)
 		if err != nil {
 			return nil, err
 		}
 	}
 
-	return so, nil
+	return options, nil
 }
 
 // Option function for setting options when configuring the service
-type Option func(*sdk.ServiceOptions) error
+type Option func(*Options) error
 
 // WithAuthToken sets a specific account token to use
 func WithAuthToken(token string) Option {
-	return func(s *sdk.ServiceOptions) error {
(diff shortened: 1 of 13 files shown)


go/services/services.go
```diff
@@ -47,76 +55,83 @@ func WithAuthToken(token string) Option {
...
-		s.ServerPort = serviceOptions.ServerPort
-		s.ServerEndpoint = serviceOptions.ServerEndpoint
-		s.ServerUseTls = serviceOptions.ServerUseTls
-		s.DefaultEcosystem = serviceOptions.DefaultEcosystem
+	return func(s *Options) error {
+		s.ServiceOptions.AuthToken = serviceOptions.AuthToken
+		s.ServiceOptions.ServerPort = serviceOptions.ServerPort
+		s.ServiceOptions.ServerEndpoint = serviceOptions.ServerEndpoint
+		s.ServiceOptions.ServerUseTls = serviceOptions.ServerUseTls
+		s.ServiceOptions.DefaultEcosystem = serviceOptions.DefaultEcosystem
```
fundthmcalculus, line 70: pointer assignment the entire structure so we don't have to update this everytime `ServiceOptions` changes?

go/services/services.go
```diff
@@ -47,76 +55,83 @@ func WithAuthToken(token string) Option {
...
+	return func(s *Options) error {
+		s.ServiceOptions.ServerEndpoint = "prod.trinsic.cloud"
+		s.ServiceOptions.ServerPort = 443
+		s.ServiceOptions.ServerUseTls = true
 
 		return nil
 	}
 }
 
 // WithProductionEnv will configure the server to use the trinsic production environment
```
fundthmcalculus:
```suggestion
// WithTestEnv will configure the server to use the environment provided by environment variables, falling back to trinsic production environment if necessary
```

fundthmcalculus (approved):
Looks good!

go/services/services.go
sethjback: This is a pointer in the `Options` struct - this function is just taking the provide options and setting them in the existing `Options`

go/services/services.go
sethjback: fixed

PR #483, opened 2022-03-11
More go and python fixes

fundthmcalculus:


go/services/service_base.go
```diff
@@ -17,26 +17,23 @@ import (
...
 // for all the other services
 func NewServiceBase(options *Options) (Service, error) {
-	if options.Channel == nil {
-		conn, err := NewServiceConnection(options.ServiceOptions)
-		if err != nil {
-			return nil, err
-		}
-
-		options.Channel = conn
+	conn, err := NewServiceConnection(options.ServiceOptions)
```
sethjback, line 20: We've made the base class private so that access to the `channel` property is no longer accessible. Are we not allowing developers to provide a grpc connection for re-use?

go/services/service_base.go
```diff
@@ -17,26 +17,23 @@ import (
...
+	if err != nil {
+		return nil, err
 	}
 
 	return &serviceBase{
 		options:          options,
+		channel:          conn,
```
sethjback, line 27: See the comment above - moving the channel out of the options and to the service base removes the ability for users to provide via options and thus removes the ability for connection re-use. Is connection reuse a design goal?

go/services/service_base.go
tmarkovski: We removed the channel as direct dependency from service construction to simplify what we pass, but channel is still available for reuse.

go/services/service_base.go
tmarkovski: It is, of secondary importance, but not as we had it designed initially.
The flaw in our design is that when passing both `options` and `channel` to a constructor method, there's side effect and/or confusion in that if you pass a channel, the options are ignored. You can then pass a `channel` connected to one server, and options pointing to another. This causes confusion and bad user experience. We had a conversation to remove this channel option from the constructor for now, and also flatten the options object to simplify it. The SDKs should now take a single configuration struct `ServiceOptions` that is defined in `sdk/options.proto` proto and represents simple configuration properties.
We don't have strict decision not to support channel reuse, nor a strict decision to support it. Goal is to simplify first.

go/services/service_base.go
sethjback: Sounds good - the way the `WithChannel` option was handled was to use the channel if provided, and if not we would create a default one.

I understand there could be some confusion if a developer provided both a grpc connection and a "connection config", since we use the provided connection and ignore the config. That being said, I would think a developer who needed to provide a custom grpc connection would not be confused when that was used (the expected behaviour) vs. the config they provided (if they did).

My confusion was just around removing the `WithChannel` option - since the base implementations are now private channel reuse in golang is no longer possible (you can get the channel, but you cannot provide that to another service when constructing it).

sethjback (approved):

PR #492, opened 2022-03-18
[dotnet] Experimental platform-specific token provider with dependency injection support

tmarkovski:
- Adds support for automatically saving and retrieving tokens for different platforms
- Adds extensions for dependency injection using common net core services
- Some cleanup in the dotnet code
- Minor changes to point tests to staging by default

The extensions are added side by side, and the previous API is unchaged


dotnet/Trinsic/AccountService.cs
```diff
@@ -35,16 +44,12 @@ public AccountService(ServiceOptions options)
...
+        if (string.IsNullOrWhiteSpace(request.EcosystemId)) request.EcosystemId = Options.DefaultEcosystem;
         var response = await Client.SignInAsync(request);
 
         var authToken = Convert.ToBase64String(response.Profile.ToByteArray());
-        
-        if (!response.Profile.Protection?.Enabled ?? true) {
-            Options.AuthToken = authToken;
-        }
+
+        if (!response.Profile.Protection?.Enabled ?? true) await TokenProvider.SaveAsync(authToken);
```
sethjback, line 52: I like that this saves the token as a default name and caches it in the concrete implementations - do we want to standardize on a naming scheme? Something like `<ecosystemID>.<email>`? though there might not always be an email.

Everything gets saved as the default name now, correct? How will users switch to use another token?

dotnet/Trinsic/Trinsic.csproj
```diff
@@ -42,23 +44,26 @@
         <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleToAttribute">
             <_Parameter1>Tests</_Parameter1>
         </AssemblyAttribute>
+        <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleToAttribute">
+            <_Parameter1>Trinsic.Browser</_Parameter1>
```
fundthmcalculus, line 48: can you put two `<_Parameter>` tags under the same `<AssemblyAttribute>` block?

dotnet/Trinsic/Storage/FileTokenProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+
+internal class FileTokenProvider : ITokenProvider
+{
+    private const string Vendor = "Trinsic";
+
+    public static FileTokenProvider StaticInstance => new();
+
+    private string? _cachedToken;
+
+    public async Task<string?> GetAsync(string name = TokenDefaults.Name, CancellationToken cancellationToken = default) {
```
fundthmcalculus, line 16: Which of these providers do we want to replicate in other languages? @sethjback  @tmarkovski do we want to create a `discussion` issue for this?

fundthmcalculus (approved):

dotnet/Trinsic/AccountService.cs
tmarkovski: The token provider is completely ignored if they pass auth token through the options, this is how they would use multi-user setups.
As you pointed out, the implementation does leave room for named tokens, once we have a clear path on how to do this. I'm curious to dig further on how many use cases there are where multi user tokens would be needed, outside of our unit tests and dev work.

dotnet/Trinsic/Trinsic.csproj
tmarkovski: No, it has to be separate, but you can have the attribute multiple times. However, I removed this specific project, but forgot to clean this up. Thanks for pointing to it.

dotnet/Trinsic/Storage/FileTokenProvider.cs
tmarkovski: Yeah, definitely start a separate discussion.

dotnet/Trinsic/AccountService.cs
fundthmcalculus: I think named tokens will be very useful in the future when developers are integrating multiple trinsic ecosystems together. Having an easy way to work with named tokens now will just be icing on the cake in the future. Also, don't forget our unit tests and dev work (as you mentioned).

I am also envisioning a future in which we are developing the sample applications in conjunction with customers @michaeldboyd  can back me up on this, and being able to switch between our ecosystem and our customer's ecosystem would be really handy.

dotnet/Trinsic/Storage/FileTokenProvider.cs
fundthmcalculus: https://github.com/trinsic-id/sdk/issues/493

PR #502, opened 2022-03-22
Sdk 479 .NET and Java/Kotlin request/response and status exceptions

fundthmcalculus:


tmarkovski (changes requested):
Can you please update the tests, so they compile and pass?

fundthmcalculus:
> Can you please update the tests, so they compile and pass?

`<joke>`Nah, I like a casual test failure. LOL.`</joke>` Obviously I'll fix that. I wanted to get your opinion before I do all the other languages similarly.

tmarkovski:
I wouldn't put too much work into ensuring the response status is properly handled like `if (response.Status != ResponseStatus.Success)`, as we will most likely change this and roll it into the grpc status code instead, which means the usual server error pipeline.
The rest of the cleanup is ok.

fundthmcalculus:
> I wouldn't put too much work into ensuring the response status is properly handled like `if (response.Status != ResponseStatus.Success)`, as we will most likely change this and roll it into the grpc status code instead, which means the usual server error pipeline. The rest of the cleanup is ok.

This was something I kind of thought should be handled at the server level. I won't worry about it for the other languages then.

dotnet/Tests/Tests.cs
```diff
@@ -110,7 +110,7 @@ public class Tests
         // Set active profile to 'allison' so we can manage her cloud wallet
         walletService.Options.AuthToken = credentialsService.Options.AuthToken = allison;
 
-        var itemId = await walletService.InsertItemAsync(new() {ItemJson = credential.SignedDocumentJson});
+        var itemId = (await walletService.InsertItemAsync(new() {ItemJson = credential.SignedDocumentJson})).ItemId;
```
tmarkovski: Does this code block go in the documentation? If yes, we should reformat it to better syntax, separate the response, etc.

dotnet/Tests/Tests.cs
```diff
@@ -327,7 +328,7 @@ public class Tests
         jsonDocument.Should().Contain(x => x.Name == "id");
         jsonDocument.Should().Contain(x => x.Name == "credentialSubject");
 
-        var itemId = await walletService.InsertItemAsync(new() {ItemJson = credentialJson.DocumentJson});
+        var itemId = (await walletService.InsertItemAsync(new() {ItemJson = credentialJson.DocumentJson})).ItemId;
```
tmarkovski: Same comment as above

dotnet/Trinsic/WalletService.cs
```diff
@@ -28,57 +28,51 @@ internal WalletService(ITokenProvider tokenProvider, IOptions<ServiceOptions> op
...
     /// </summary>
     /// <remarks>
-    /// See https://docs.microsoft.com/en-us/azure/cosmos-db/sql-query-select
+    ///     See https://docs.microsoft.com/en-us/azure/cosmos-db/sql-query-select
     /// </remarks>
     /// <returns></returns>
     public async Task<SearchResponse> SearchAsync(SearchRequest request) {
-        if (String.IsNullOrWhiteSpace(request.Query))
+        if (string.IsNullOrWhiteSpace(request.Query))
             request.Query = "SELECT * FROM c";
```
tmarkovski: Can we change this to `SELECT c.id, c.type, c.data FROM c`? It's a cleaned up version.

tmarkovski (changes requested):

dotnet/Tests/Tests.cs
fundthmcalculus: It does, but this is pretty self-explanatory as it is.

dotnet/Trinsic/WalletService.cs
fundthmcalculus: I can make this change everywhere.

dotnet/Tests/Tests.cs
tmarkovski: I know, it's just a bad C# code that I'm not too proud of us suggesting in a docs sample

dotnet/Trinsic/WalletService.cs
tmarkovski: This only applies to Wallet search, not Template or TrustRegistry searches

fundthmcalculus: General issue here to prevent off-topic or loss: https://github.com/trinsic-id/sdk/issues/504

dotnet/Tests/Tests.cs
```diff
@@ -110,7 +110,8 @@ public class Tests
         // Set active profile to 'allison' so we can manage her cloud wallet
         walletService.Options.AuthToken = credentialsService.Options.AuthToken = allison;
 
-        var itemId = await walletService.InsertItemAsync(new() {ItemJson = credential.SignedDocumentJson});
+        var insertItemResponse = (await walletService.InsertItemAsync(new() {ItemJson = credential.SignedDocumentJson}));
```
tmarkovski:
```suggestion
        var insertItemResponse = await walletService.InsertItemAsync(new() {ItemJson = credential.SignedDocumentJson});
```

dotnet/Tests/Tests.cs
```diff
@@ -327,7 +329,8 @@ public class Tests
         jsonDocument.Should().Contain(x => x.Name == "id");
         jsonDocument.Should().Contain(x => x.Name == "credentialSubject");
 
-        var itemId = await walletService.InsertItemAsync(new() {ItemJson = credentialJson.DocumentJson});
+        var insertItemResponse = (await walletService.InsertItemAsync(new() {ItemJson = credentialJson.DocumentJson}));
```
tmarkovski:
```suggestion
        var insertItemResponse = await walletService.InsertItemAsync(new() {ItemJson = credentialJson.DocumentJson});
```

tmarkovski (approved):

PR #503, opened 2022-03-23
[dotnet] use base64url encoding everywhere

tmarkovski:
Ensures dotnet uses base64 url encoding and decoding


fundthmcalculus (approved):
Don't worry about the other languages, I already did them in base64-url encoding

PR #506, opened 2022-03-24
Ruby Refactor and Demo Extraction with `Request` `Response` structure update

### LINKED ISSUE #479: Ensure every language uses the Request and Response structure objects
This will allow us to iterate the contents of the structure without iterating the content of the SDK wrappers. For endpoints that return status, throw an appropriate exception instead of returning a failure status.

Languages:
- [x] .NET
- [x] Go
- [x] Java
- [x] Node
- [x] Python
- [x] Ruby
- [x] Swift
- [x] Web

fundthmcalculus:


fundthmcalculus:
This works on staging on my machine, the github action runner is acting up.

ruby/lib/services/account_service.rb
```diff
@@ -0,0 +1,57 @@
+require 'services/service_base'
+
+module Trinsic
+  class AccountService < ServiceBase
+
+    def initialize(service_options)
```
tmarkovski: Can `service_options` be made nullable or overloaded so that it's not required to be passed?

ruby/lib/services/credential_service.rb
```diff
@@ -0,0 +1,57 @@
...
+      if @service_options.server_use_tls
+        channel_creds = GRPC::Core::ChannelCredentials.new
+        @client = Credentials_V1::VerifiableCredential::Stub.new(get_url, channel_creds)
+      else
+        @client = Credentials_V1::VerifiableCredential::Stub.new(get_url, :this_channel_is_insecure)
+      end
+    end
+
+    def issue_credential(document)
+      request = Credentials_V1::IssueRequest.new(document_json: JSON.generate(document))
```
tmarkovski: Should this expose the request object directly?

ruby/lib/services/account_service.rb
fundthmcalculus: That will be done.

ruby/lib/services/credential_service.rb
fundthmcalculus: In the next PR, it will. I'm separating the refactor from the `request`/`response` work for git history sake. (file rename)

tmarkovski (approved):

PR #542, opened 2022-04-06
Document Account Service Request/Response Models

lucasamonrc:
Closes Issue #515

### DIFF
--- a/docs/reference/services/account-service.md
+++ b/docs/reference/services/account-service.md
@@ -2,13 +2,8 @@
 
 ### Construct Instance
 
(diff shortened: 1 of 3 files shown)


lucasamonrc:
One thing that particularly hard when working on this, is how should these docs be formatted. What should be included and what shouldn't. 

What I am looking for as other review this PR, is feedback on what needed to be included and what shouldn't, that way I can create a solid framework I can use when writing documentation for other services.

fundthmcalculus (approved):

PR #546, opened 2022-04-06
git docs authors plugin

fundthmcalculus:
Author plugin for `mkdocs`. It adds the author to the bottom of every page, as shown here:
image

### DIFF
--- a/docs/overrides/main.html
+++ b/docs/overrides/main.html
@@ -3,13 +3,13 @@
 {% block content %}
   {{ super() }}
 
-  <!-- {% if git_page_authors %} -->
+  {% if git_page_authors %}
     <div class="md-source-date">
       <small>
           Authors: {{ git_page_authors | default('enable mkdocs-git-authors-plugin') }}
       </small>
     </div>
-  <!-- {% endif %} -->
+  {% endif %}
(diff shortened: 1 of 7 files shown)


tmarkovski:
@fundthmcalculus Mind posting a screenshot how this looks when run locally?

fundthmcalculus:
image
Here is an example - alphabetical order.

docs/overrides/main.html
```diff
@@ -0,0 +1,17 @@
+{% extends "base.html" %}
+
+{% block content %}
+  {{ super() }}
+
+  <!-- {% if git_page_authors %} -->
```
geel9: What's the rationale behind commenting this out? Will this mean that we'll have a blank `Authors: ` entry on pages that have been `exclude`d in the plugin config?

geel9 (changes requested):
With the change moving the authors to the bottom, and assuming a positive resolution of the one comment I have, looks good to me.

docs/overrides/main.html
fundthmcalculus: good catch, that was an experiment on my end. reverted.

geel9 (approved):

PR #558, opened 2022-04-07
Clarify docs surrounding wallet architecture.

d1str0:
Previous text was a bit fragmented and confusing. I hope my proposal is still accurate.


fundthmcalculus (approved):
Thank you!

PR #569, opened 2022-04-13
Inject all go samples

fundthmcalculus:


go/services/credentialtemplate_service_test.go
```diff
@@ -102,4 +77,35 @@ func TestTemplatesDemo(t *testing.T) {
 	assert2.NotNil(jsonDocument)
 	assert2.NotNil(jsonDocument["id"])
 	assert2.NotNil(jsonDocument["credentialSubject"])
+
+	// getCredentialTemplate() {
+	getResponse, err := templateService.Get(context.Background(), &sdk.GetCredentialTemplateRequest{Id: template.Data.Id})
+	// }
+	if getResponse != nil {
```
sethjback, line 84: Here I would normally expect to check for `err` being nil and if not assume the call succeeded (and that should mean that `getResponse` is not nil)

fundthmcalculus: this is for docs, we don't care the response, and go complains if the variable is unused.

sethjback: Ahhh, missed the `//` so the `getResponse` check isn't injected into the docs. :+1:

sethjback (approved):

PR #571, opened 2022-04-14
`provider-service.md` and `template-service.md` is complete

fundthmcalculus:


geel9 (changes requested):
The tabs are broken on both pages:

image

(Selecting Go in this example causes Java and Ruby to appear below; selecting any of the preceding languages causes Java and Ruby to disappear)

To fix this, you need to add a newline before each of the language tabs which are appearing erroneously -- before Java and Ruby in this example.

geel9 (approved):

PR #578, opened 2022-04-14
fix unit test bugs

fundthmcalculus:


python/samples/provider_demo.py
```diff
@@ -8,22 +8,20 @@
...
     account_service = AccountService(server_config=trinsic_config())
-    account_profile, _ = await account_service.sign_in()
+    account_profile = await account_service.sign_in()
     provider_service = ProviderService(server_config=trinsic_config(account_profile))
-    invite_response = await provider_service.invite_participant(
-        request=InviteRequest(
-            details=AccountDetails(email="nothing@trinsic.id"),
-            participant=ParticipantType.participant_type_individual,
-            description="I dunno",
+    try:
```
geel9, line 14: Doesn't exactly seem like a _fix_ so much as a _suppression_ but I'll trust that you know what you're doing

geel9 (approved):

python/samples/provider_demo.py
fundthmcalculus: This is a known failure because the account doesn't exist, but we use this code for docs/samples. My longer term plan is to have the doc-injected code separated, since it doesn't really add to the test cases in the same way.

PR #588, opened 2022-04-19
Macro for Protobuf Message injection in docs

geel9:
Currently, to include documentation for a protobuf message in a page, the following syntax must be used:

`{{ include_section('reference/proto/', 'ServiceOptions') }}`

Which produces the following result:

This PR introduces a new macro, `proto_obj()`, which can be used to quickly inject documentation for any Protobuf message. Its usage is as such:

`{{ proto_obj('ServiceOptions') }}`

Which produces the following result:

This has the following benefits:

- Simpler to include a protobuf object; they are treated as a first-class citizen
- Automatically includes a link to the object definition on the protobuf documentation page
- Packages the object's name, its description, and its fields into a visually parseable block, clearly separated from the rest of the page's contents
- Produces HTML that can easily be styled by us later on
- Removes annoying "Top" link that is injected from the protobuf documentation page (which doesn't really belong in pages it's injected into)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -21,6 +21,34 @@ The docs are easy to edit. Find the `nav` tag in [./mkdocs.yml](.mkdocs.yml) fil
 
 Write markdown in the `docs` folder. You can use normal markdown, html, or any of our theme's features - [Reference](https://squidfunk.github.io/mkdocs-material-insiders/reference/abbreviations/)
 
+### Trinsic-specific Macros
+
+To make editing docs easier to work on (and more powerful), we make use of a few macros that we've developed -- the source for these lives in the `mkdocs_macros_main.py` file in the root of this repository.
+
+#### `include_section`
+
+This macro includes a section from another markdown file.
+
+Usage: `{{ include_section(file_name: str, section_name: str, include_heading: bool=False) }}`
+
+- `file_name` is the path, _relative to the docs root_, of the markdown file you wish to include a section from
+- `section_name` is the name of the section you wish to include
+- `include_heading` controls whether or not to include the heading itself
+
+#### `proto_obj`
+
+This macro allows you to inject the documentation for any protobuf message or service that is documented in `reference/proto/index.md`.
+
+Usage: `{{ proto_obj(obj: str, header_text: str=None) }}`
+
+- `obj` is the name of the protobuf message or service
+    - For example, `ServiceOptions`
+- `header_text`, if specified, overrides the text of the header with whatever you specify
+    - If unspecified, the name of the message is used (same as the value of `obj`)
+
+
+Example: `{{ proto_obj('SignInRequest') }}`
+
 ### Build and deploy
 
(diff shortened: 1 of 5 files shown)


michaeldboyd:
Thanks! This looks super useful. Where can a new dev go to understand how to use this macro? Might we want to add a section to the docs readme about this?

geel9:
> Thanks! This looks super useful. Where can a new dev go to understand how to use this macro? Might we want to add a section to the docs readme about this?

Good point; I'll work on an addition to the readme which goes over things like that.

fundthmcalculus (commented):
My one concern with this approach is that it forces all the formatting on us. We can't take advantage of markdown formatting. We should really be exporting markdown formatting not HTML directly

fundthmcalculus:
> Thanks! This looks super useful. Where can a new dev go to understand how to use this macro? Might we want to add a section to the docs readme about this?

Also add a section for the other macro, which enables section includes directly, independently of it being a proto generated `.md` file. My goal is to eventually move most of this work upstream of us into an mkdocs plugin directly.

fundthmcalculus (approved):

geel9:
> My one concern with this approach is that it forces all the formatting on us. We can't take advantage of markdown formatting. We should really be exporting markdown formatting not HTML directly

So, my philosophy diverges from yours a bit here.

In my opinion, when it comes to this sort of thing, markdown should be emitted by _people_, and HTML should be emitted by _tools_. I think Markdown is nice, because it reduces the human difficulty of visually parsing out a document and making modifications (compared to HTML), but in doing so it makes a number of tradeoffs. 

Fundamentally, this macro is emitting a structured element, which we want control over -- it seems to me that emitting HTML from the macro, and then styling it, fits this purpose. (*Also, because whitespace is semantic in markdown far moreso than HTML, doing raw injection of markdown into markdown can get messy -- for example, aligning indentation.*)

The way I see it, the alternatives to emitting HTML from the macro were thus:
- Create a Markdown extension purpose-built for this
  - A more complete / in-toolchain solution
  - More overhead to get set up; still requires us to write code that emits basically the same HTML somewhere in the toolchain.
- Reuse an existing Markdown extension
  - Less control
  - Less effort to get going
- Reuse infoboxes / blockquotes / etc.
  - Least control
  - Least effort
  - Looks... amateur.


Ultimately, in order to have world-class (let's be honest... Stripe-class) documentation, I think we're going to have to be building a much more bespoke solution -- building our own documentation site (or at least, a heavily-modified one from a base). 

Five years from now, I don't imagine any of our current documentation (or tooling related to the documentation) will even exist. If we're doing automatic inclusion of content at all (say, of protobuf files), I would hope we have the engineering bandwidth to do so in a less... messy way.

I do see where you're coming from, and I do want to keep things sane. Having _humans_ write HTML inside markdown is definitely a path to insanity (barring certain one-off situations), but I think writing tools to do so is necessary to have documentation that looks and feels wonderful.

PR #591, opened 2022-04-20
[sc-9197] sdk/go: added ability to configure client-side middleware

dududko:
Added ability to configure middlewares for go grpc client. 
This will enable sdk consumers to configure monitoring, logging, tracing or other [grpc middlewares](https://github.com/grpc-ecosystem/go-grpc-middleware).


fundthmcalculus:
Overall, I like this! Let us figure out how we're planning to handle such things across languages (this is a cross-cutting concern, and I don't want to duplicate efforts or end up with incompatible implementations, if possible).

fundthmcalculus (approved):

PR #595, opened 2022-04-21
Enable navigation tabs in docs top level

tmarkovski:


mkdocs.yml
```diff
@@ -55,6 +55,7 @@ theme:
     - navigation.instant
     - navigation.tracking
     - announce.dismiss
+    - navigation.tabs
```
fundthmcalculus, line 58: This breaks the layout that we've been using and makes it more confusing. @michaeldboyd requested this be reverted last time.

fundthmcalculus (changes requested):

mkdocs.yml
tmarkovski: @michaeldboyd Can you please shed more light on what layout this breaks? UX wise, top level navigation tabs are a lot more usable and discoverable. It also allows us to go one level deeper in the left hand navigation menu. It also separates the sections conceptually a lot better than having them be bundled in one menu.

geel9: I'm a +1 on top-level tabs

lucasamonrc: I like top-level tabs as well. But I also like to be able to navigate using the sidebar when I am reading docs. So why not both?

lucasamonrc: I like top-level tabs as well. But I also like to be able to navigate using the sidebar when I am reading docs. So why not both?

geel9: @lucasamonrc The way I envision it would be to have the top-level docs be the root navigation nodes, and the sidebar would be for the navigation nodes that follow from whatever top-level tab you're browsing.

So, if we had "SDK" and "Learn" tabs, the sidebar would be entirely different for each of those tabs.

I think we just have too many things to write about, and we can only feasibly nest the sidebar to a certain extent. Adding the top-level tabs basically gives us one more nesting level, as well as a clean mental (and actual) partition between logical sections.

Our SDK reference doesn't need to share space with our conceptual documentation, IMO. Right now, they're in competition, by virtue of the aforementioned limit of tree depth for the sidebar.

I think right now we're waiting for @michaeldboyd to provide input?

fundthmcalculus (approved):

fundthmcalculus:
Let's give this a shot.

lucasamonrc (approved):

PR #599, opened 2022-04-21
Document Templates Service Request/Response Models

### LINKED ISSUE #516: Document Templates Service Request/Response Models
For each endpoint call, document the following things:

### Description
Ideally coming from the proto file, this description should include a high level explanation of the functionality
How might we import this content directly from the proto file's service?

### Request and Response Data Models
Links to the data models for the request and response bodies should be included. Just the links to the data models are needed within the reference markdown file, not the full tables of the data models - but that would be cool.

lucasamonrc:
Closes #516

### DIFF
--- a/docs/reference/services/template-service.md
+++ b/docs/reference/services/template-service.md
@@ -1,25 +1,31 @@
 
-# Credential Templates
+# Template Service
 
-Credential Templates are json templates that help issuers save context when issuing credentials. 
+Credential Templates are JSON templates that help issuers save context when issuing credentials. 
 
 When issuing a credential through Trinsic, it's as simple as adding the attributes to the credential template and then issuing it. 
 
 These credential templates can be shared between issuers in an ecosystem. 
 
-A template is a saved JSON-LD schema that is used to define/create credentials and verification. 
+## Create Template
 
-### Create Credential Template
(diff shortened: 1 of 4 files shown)


docs/reference/services/template-service.md
```diff
@@ -149,26 +164,20 @@ A template is a saved JSON-LD schema that is used to define/create credentials a
...
-    <!--codeinclude-->
-    ```typescript
-    
-    ```
-    <!--/codeinclude-->
+When using one of the SDKs, you must supply a [Delete Credential Template Request](../proto/index.md#deletecredentialtemplaterequest) object. This object follows the model below:
+
+{{ proto_obj('DeleteCredentialTemplateRequest') }}
+
+Then you can supply it to SDK:
```
fundthmcalculus, line 180: You can't split code tab sections. Move this above the "Trinsic CLI" section

fundthmcalculus (changes requested):

docs/reference/services/template-service.md
lucasamonrc: I can't because it is not possible, or I can't because I shouldn't? I am just confused by this, since every reference page so far I have been doing this, but only now it seemed to be a problem.

fundthmcalculus (approved):

tmarkovski (approved):

PR #610, opened 2022-04-26
Move web tests to typescript: jasmine, karma

fundthmcalculus:
This allows us to test web with either the console or the web browser. Hypothetically, this could enable running `@trinsic/web` via node on the server.


tmarkovski (approved):
Great work!
As part of the next release, we should verify that this works with webpack and possibly the base react sample. Those are our manual smoke tests when it comes to integration.

fundthmcalculus:
> Great work! As part of the next release, we should verify that this works with webpack and possibly the base react sample. Those are our manual smoke tests when it comes to integration.

This is something @MichaelEdwardBlack and I are already looking into. Good points on testing there. Karma tests here run under webpack already.

PR #611, opened 2022-04-27
changed to base64 encoding

MichaelEdwardBlack:
browser buffer doesn't have support for base64url encoding so I changed it to base64. See reference here https://github.com/feross/buffer/blob/795bbb5bda1b39f1370ebd784bea6107b087e3a7/index.js#L379


fundthmcalculus:
[Use this instead](https://www.npmjs.com/package/base64url). It will keep typescript consistent with every other language where we use `base64url` encoding. In my opinion, `base64url` should have been the default, lol.

fundthmcalculus:
No longer necessary, closing.

PR #622, opened 2022-05-04
Build protobuf documentation reference during docs release action

geel9:
Modifies the `release-docs` GitHub action to build the Protobuf reference markdown file before building and shipping the docs.

Without this step, said Protobuf Reference will only be updated when a developer manually executes `devops/generate_proto_files.py` and pushes the file to the repo.

Thoughts:
- Maybe this should only happen when we release a new tagged version of the SDK?
  - Ensures that newer proto files, which haven't been released in the SDK yet, are not documented before they should be.


fundthmcalculus (commented):
add `pip install -r ./devops/requirements.txt` to the previous run action, and this should be good.

PR #628, opened 2022-05-04
template service count parameter unused

### LINKED ISSUE #583: SearchResponse Count is 0, although there are Items in the response
### Reproduction steps

1. Create a credential and insert it into a wallet
2. From the `WalletService` call `SearchAsync`
3. The `SearchResponse` has a property named `Count` which is `0` even though the Items list has 1 item.
4. `Items.Count` is equal to `1`

image

Could it be that the `Count` on the `SearchResponse` references a different count? In that case it is a bit ambigious IMHO.

### Reproduction Code

_No response_

fundthmcalculus:
closes #583


proto/services/verifiable-credentials/templates/v1/templates.proto
```diff
@@ -21,7 +21,7 @@ message SearchCredentialTemplatesRequest {
 message SearchCredentialTemplatesResponse {
   string items_json = 1;
   bool has_more = 2;
-  int32 count = 3;
```
tmarkovski, line 24: The source of truth for these proto files is the server repo. Please make sure to add this change there, otherwise it'll get overwritten next time we copy the changes from there

tmarkovski (approved):

proto/services/verifiable-credentials/templates/v1/templates.proto
fundthmcalculus: [see here](https://github.com/trinsic-id/server/pull/566)

PR #632, opened 2022-05-10
Documentation Navigation Reorganization

geel9:
This PR reorganizes the navigation tree to provide a cleaner separation between categories and reduce noise in the navbar.

- `Home` now contains `Join the Beta` (which was previously `Before You Start`)
- `Learn`, a new section, combines `Getting Started` and `Key Concepts`
- `SDK Reference` has been renamed to `Services`, and now comes after `Learn`
  - This will allow us to document all our services, such as OIDC, in this section.
- `Developer Tools` has been renamed to `Tools & SDKs`

**Notes**

- `learn/index.md` needs to be iterated upon; it is very barebones right now
- We need to figure out how we will distinguish our interactive demos (Katacoda) from our non-interactive tutorials.
- The actual locations of each markdown file has not been changed; URLs will be somewhat detached from the navigation tree
  - Once we are sure we have a "final" layout we can move the files around and fix up all references to them.


mkdocs.yml
```diff
@@ -5,19 +5,32 @@ edit_uri: edit/main/docs/
...
+  - Services:
+    - Overview: reference/index.md
+    - SDK Services:
+      - Account Service: reference/services/account-service.md
+      - Credential Service: reference/services/credential-service.md
+      - Provider Service: reference/services/provider-service.md
+      - Template Service: reference/services/template-service.md
+      - Trust Registry Service: reference/services/trust-registry-service.md
+      - Wallet Service: reference/services/wallet-service.md
+      - Protocol Buffers Reference: reference/proto/index.md
```
fundthmcalculus, line 32: Maybe move protocol buffers reference over to `Resources`? It's not a service, just a reference.

fundthmcalculus (approved):

mkdocs.yml
geel9: I'm of split minds on it -- it'd probably fit well in either place.

On the one hand, it's inherently related to the SDK Services, and it makes sense to live alongside them as a reference; on the other hand, I see how this is slightly mixing concerns.

I'll leave it where it is for now, but we can come back to it.

fundthmcalculus: Same. Compromise and hyperlink from the other plausible location?

fundthmcalculus: and/or rename it from `Protocol Buffers Reference` to `Service Objects Protocol Buffers Reference` ? That way it's clear why it belongs here. It's the reference list of objects you can pass to our services.

PR #634, opened 2022-05-10
Add OIDC documentation first pass

### LINKED ISSUE #613: OIDC4VP Documentation
- [x] Reorganize documentation site
  - Must be done first to provide a good place for OIDC4VP docs to live 
- [x] Create documentation for integrating with our OIDC4VP service

geel9:
This PR adds documentation for the OIDC Service, under `Learn -> Other Services -> OpenID Credential Exchange`

Resolves #613 

**Notes**
- This page does not explain the OIDC flow itself -- rather, it simply tells users to use an OIDC library for their language of choice.
- This is a first pass, intended to make the information available.

**Future Work**
- Iterate on this page
  - Improve `Configuration` section
  - Add more detail to `Response Data` section
  - Maybe detail OIDC flow; maybe leave that out of scope. I'm not sure on this one.
- Explain the intricacies of server-driven vs. client-driven flow
  - Instruct users that client-driven has limited trust
  - Go over how to implement server-driven appropriately, through proper session management and use of `code_verifier/code_challenge` and `state`.
  - Even if we don't document the OIDC Protocol itself, I think we should still cover this specific area. It's important that we encourage healthy and secure patterns of implementation.
- Create OIDC Tutorial
  - Using a client library, create a basic implementation against this service from scratch.

### DIFF
--- a/docs/reference/other/openid.md
+++ b/docs/reference/other/openid.md
@@ -1,0 +1,113 @@
+# OpenID Credential Exchange Service
+
+## Overview
+Trinsic provides an [OpenID Connect <small>:material-open-in-new:</small>](https://openid.net/connect/){target=_blank} ("OIDC") service which enables verifiers to request credentials from a user's cloud wallet in a simple and secure way.
+
+This service does not require the use of Trinsic's SDKs in the user's browser, and is therefore lightweight and easy to integrate.
+
+!!! info "Credential Issuance"
+    Currently, Trinsic's OpenID Connect service only enables the sharing of a credential between holder and verifier. We plan to support issuance through this service at a later date.
+
+    In the meantime, use the [InsertItem](../services/wallet-service.md#insert-item) SDK call to store credentials in a holder's wallet.
+
+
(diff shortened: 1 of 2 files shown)


docs/reference/other/openid.md
```diff
@@ -0,0 +1,113 @@
...
+    Note that the above data has been modified for brevity. 
+
+    The `@context` and `type` arrays will contain additional entries which are specific to the credential.
+
+
+### Verify the Received Proof
+
+`vp_token` is a Verifiable Proof; before making use of its data, you must verify it. 
+
+This is as simple as [passing the proof to the Verify Proof call](../../services/credential-service/#verify-proof).
```
tmarkovski:
```suggestion
This proof can be verified with any library that supports VC verifications for BBS+ signatures. It can also be verified using Transit's SDK; this is as simple as [passing the proof to the Verify Proof call](../../services/credential-service/#verify-proof).
```

tmarkovski (approved):

PR #637, opened 2022-05-11
CLI update with Trust Registry commands and services

tmarkovski:


cli/build.rs
```diff
@@ -40,34 +35,14 @@ fn main() {
...
-    move_file!(
-        "./src/proto/services.account.v1.rs",
-        "./src/proto/services/account/v1/mod.rs"
-    );
-    move_file!(
-        "./src/proto/services.trustregistry.v1.rs",
-        "./src/proto/services/trustregistry/v1/mod.rs"
-    );
+    move_file!("./src/proto/google.protobuf.rs", "./src/proto/google/protobuf/mod.rs");
+    move_file!("./src/proto/services.options.rs", "./src/proto/services/options.rs");
```
geel9: I suspect the path to `options` may be incorrect?

The path of `options.proto` is `services/options/options.proto`; this snippet seems to assume an incorrect path of `services/options.proto`

tmarkovski: Actually it's an empty file that isn't referenced. The `prost` runtime doesn't support field extensions, it just parses them into UnknownOption, so we can safely ignore this file.

PR #673, opened 2022-05-25
Vaccine Walkthrough C# Test Suite & Serialization Rewording

geel9:
- Vaccine walkthrough snippets for .NET are now injected from `dotnet/Tests/Samples/VaccineWalkthroughTests.cs`
- Removed sample code demonstrating the saving of account profiles to disk
  - They're just strings; people can figure that out on their own
  - Made more sense when they were actual objects


fundthmcalculus (approved):
Are we duplicating the vaccine demo test now, or is it just moved? I'd rather not duplicate tests more than necessary.

PR #686, opened 2022-05-27
Java Vaccine Walkthrough Fixes

geel9:
**Main Changes**
- Created `CredentialsDemo.java` and `WalletsDemo.java`
  - `credential-service.md` and `wallet-service.md` now inject from these files instead of `VaccineDemo.java`
- Updated `VaccineDemo.java` in line with new changes

**Misc Changes**
- Removed line from `templates-service.md` indicating that the Verifiable Presentation Request Specification would be usable to query templates in future
  - Seems like an oversight -- likely copied over from the same section in `wallet-service.md`
- Added `serviceBase.setDefaultEcosystem()` method


java/src/main/java/trinsic/services/ServiceBase.java
```diff
@@ -14,6 +14,7 @@
 import trinsic.services.account.v1.AccountOuterClass;
 
 import java.util.Base64;
+import java.util.Optional;
```
fundthmcalculus: what is this for?

fundthmcalculus (approved):

java/src/main/java/trinsic/services/ServiceBase.java
geel9: No idea... must've been the ol' spicy IDE auto-import.

Will fix.

fundthmcalculus (commented):
commit comment checks out.

PR #703, opened 2022-06-03
Update protos from server with changes to account API

tmarkovski:
Proto update related to the account login flow. Dependent on the server PR


fundthmcalculus (changes requested):
Blocking merge until I generate the proto bindings for all languages. Will be a PR into this branch.

Otherwise, I approve.

tmarkovski:
An issue it seems with the typescript wrapper due to exporting types with the same name. Both Account and Provider have `InfoRequest` and `InfoResponse`, even though this is legal in proto, in our TS library all are exported under the same name object.
We'll need to resolve this case.

sethjback:
This namespace clashing was an issue in golang as well - ideally we will want to not modify the generated proto files in a build step since it can cause issues with what each library is expecting

fundthmcalculus:
> This namespace clashing was an issue in golang as well - ideally we will want to not modify the generated proto files in a build step since it can cause issues with what each library is expecting

Agreed, but we had to do so due to some conflicts - off the top of my head, I don't remember what it was.



> An issue it seems with the typescript wrapper due to exporting types with the same name. Both Account and Provider have `InfoRequest` and `InfoResponse`, even though this is legal in proto, in our TS library all are exported under the same name object. We'll need to resolve this case.

We barrel export everything, which is an issue. That said, we have to export it so that our users have access to it. Let me see if we can namespace export.

tmarkovski:
We can also rename the messages to be globally unique. This shouldn't be a big deal given the number of API endpoints. `AccountInfoRequest` and `EcosystemInfoRequest` seem pretty OK.
Also, renaming the request message name is protocol backward compatible, so can be done pretty easy.
I'd rather avoid having to use namespaced types `account.InfoRequest`.
For typescript, another option is to rename the export itself, `export InfoRequest as AccountInfoRequest`.

fundthmcalculus:
> We can also rename the messages to be globally unique. This shouldn't be a big deal given the number of API endpoints. `AccountInfoRequest` and `EcosystemInfoRequest` seem pretty OK. Also, renaming the request message name is protocol backward compatible, so can be done pretty easy. I'd rather avoid having to use namespaced types `account.InfoRequest`. For typescript, another option is to rename the export itself, `export InfoRequest as AccountInfoRequest`.

I would rather not alias things at the language level where possible. I would rather change the name inside the proto, since that is broadly reapplicable.

fundthmcalculus (approved):

PR #704, opened 2022-06-03
Golang single service

sethjback:
This draft is a proposal for 2 things:

1. Fixing the package namespace issue in the protobuf. Ideally this will be pushed to a separate PR, it is included here simply to make everything work.
2. Creates a `trinsic-service` that wraps all the individual services

This is intended mostly to encourage feedback on the wrapper service


go/services/credential_service.go
```diff
@@ -2,7 +2,8 @@ package services
 
 import (
 	"context"
-	credential "github.com/trinsic-id/sdk/go/proto/verifiablecredentials/v1"
+
+	credentials "github.com/trinsic-id/sdk/go/proto/verifiablecredentials/v1"
```
fundthmcalculus: I specifically went through the languages and made all services singular, not plural. Is there a reason for this being reverted?

fundthmcalculus (approved):
In general, I like this. Except for the singular/plural issue, which might just be a personal preference.

go/services/credential_service.go
sethjback: Nope! Singular works - this is just my RESTful instinct coming out....

fundthmcalculus (changes requested):
Blocking merge until @sethjback has time to confirm I didn't break anything

sethjback:
Looks good to me

fundthmcalculus (approved):

PR #709, opened 2022-06-06
Protobuf Object Documentation Overhaul

### LINKED ISSUE #636: Refactor documentation protobuf object tooling
With #635, protobuf fields which are not required are marked with the `options.optional` attribute (FQN is `services.options.optional`).

This attribute exists solely for documentation generation -- when generating proto docs, we should label optional fields as such.

geel9:
## Overview

This PR adds a number of improvements to the way we document our protobuf objects (services, methods, messages, enums, etc.).

Presently, we use `protoc-gen-doc` to generate a static HTML page listing all protobuf objects, and then inject from that page whenever we want to document protobuf objects elsewhere.

However, this doesn't support custom options (meaning our custom `optional` option can't be represented), and it prevents us from doing anything particularly interactive or useful, such as embedding sub-type information in fields.

Resolves #636 

#### Introduction of `protoc-gen-json`

I've created [`protoc-gen-json`](https://github.com/trinsic-id/protoc-gen-json), a `protoc` plugin which exports a JSON file containing structured information about each object defined in the protobuf files that were compiled.

This is ingested by our python macros to generate HTML for any protobuf object using mkdocs macros.

#### New protobuf object design

Protobuf objects are now documented using auto-generated HTML.

If a field has a complex (non-primitive) type, the type information will be embedded as an expandable section inline. This supports multiple nesting levels.

#### New display of method request / response types

All SDK methods are now documented using a standardized format with tabs for:

- Sample Code
- Request Type
- Response Type

## To-do

- [x] Add support for enum rendering
- [x] Add `protoc-gen-json` to GitHub Actions
  - Presently, `proto.json` must be built manually and committed.
- [x] Document the new macros added / how to use this functionality as a doc writer


fundthmcalculus (approved):
WAY TO FREAKIN GO!

PR #710, opened 2022-06-06
Update golang packages

sethjback:
Modifies the SDK `proto` files to change the golang package name, keeping it consistent with the directory layout in `/proto/`.

Updates the generated proto files in `go/proto/` to use the new namespaceing.

Updates the `go/services/` files to use the new imports.

Changing the package names prevents collisions with services that have the same endpoints (e.g. `account.info` and `provider.info`) and allows for language tooling to auto-resolve packages.


fundthmcalculus (approved):

fundthmcalculus (changes requested):
Blocking merge until we verify that this is the proper set of proto files with all of @tmarkovski changes from server

tmarkovski (approved):

fundthmcalculus (approved):

tmarkovski:
Minor PR that aligned server with sdk protos - https://github.com/trinsic-id/sdk/pull/712

sethjback:
Looks good

PR #712, opened 2022-06-07
Minor proto fix

tmarkovski:


proto/services/account/v1/account.proto
```diff
@@ -171,6 +161,16 @@ message LoginConfirmResponse {
     AccountProfile profile = 1;
 }
 
+// Authorize ecosystem to receive wallet envents
```
geel9:
```suggestion
// Authorize ecosystem to receive wallet events
```

fundthmcalculus (approved):

proto/services/account/v1/account.proto
geel9: This has to happen on server first, actually.

tmarkovski: Yup

fundthmcalculus (approved):

PR #761, opened 2022-06-16
Sdk 759 go java dart python ruby

fundthmcalculus:


java/src/main/java/trinsic/services/ProviderServiceKt.kt
```diff
@@ -52,11 +55,11 @@ class ProviderServiceKt(options: Options.ServiceOptions?) : ServiceBase(options)
 
   @Throws(InvalidProtocolBufferException::class, DidException::class)
   suspend fun getOberonKey(request: GetOberonKeyRequest): GetOberonKeyResponse {
```
geel9, line 57: `GetOberonKey` does not require authentication

python/trinsic/__main__.py
```diff
@@ -5,7 +5,7 @@
 
 async def demo():
     account_service = AccountService()
-    profile = await account_service.sign_in()
+    profile = await account_service.login_anonymous()
```
geel9: Maybe this demo could be changed to `GetOberonKey` to avoid creating tons of needless accounts in prod?

python/trinsic/service_base.py
```diff
@@ -27,9 +29,11 @@ def _update_metadata(
     metadata: "_MetadataLike",
     request: "_MessageLike",
 ) -> "_MetadataLike":
-    if route in _skip_routes:
-        return metadata
     metadata = metadata or {}
+    # Remove this key
+    authenticate_call = metadata.pop("authenticateCall", False)
+    if route in _skip_routes and not authenticate_call:
```
geel9, line 35: I'm not in love with *both* `_skip_routes` and `authenticate_call` existing (ideally it'd be either of them), but I get why they do.

geel9 (approved):

java/src/main/java/trinsic/services/ProviderServiceKt.kt
fundthmcalculus: @tmarkovski where is the master list of which endpoints require authentication? We need to keep track of that.

python/trinsic/service_base.py
fundthmcalculus: `_skip_routes` is because we skip by endpoint, up until now. Now we have conditional authentication. I can remove `skip_routes`

python/trinsic/__main__.py
fundthmcalculus: the main demo isn't called as part of test. It's called manually by the user to verify installation.

geel9: Right, but even so, the ability to fetch the Oberon key would still serve the same purpose -- I don't think we need to create an account in prod to verify that the install worked successfully.

java/src/main/java/trinsic/services/ProviderServiceKt.kt
fundthmcalculus: https://github.com/trinsic-id/sdk/issues/763

PR #766, opened 2022-06-19
[cli] Add machine readable output format

### LINKED ISSUE #691: CLI: Add `--out` param to `account login`
As a user, I want to be able to log in to an account via the Trinsic CLI, and have the authentication token saved to an arbitrary file of my choosing.

This would look like:

`trinsic account login --name "Allison" --out "allison.txt"`

Because of the way the response is formatted to the console, it's non-trivial to do this in bash currently -- extracting _just_ the auth token from the formatted output is annoying.

This would also have benefits for our walkthroughs. Presently, we [awkwardly ask users to do this by hand](https://user-images.githubusercontent.com/1294419/171429443-0057864d-0f09-40b7-946e-f8217b26c6e3.png), which is annoying and disrupts the flow of the walkthrough.

tmarkovski:
Adds support for `--json` or `-j` flag to `trinsic` command which formats all output as JSON without color, suitable for processing with tools, like `jq` or `ConvertFrom-Json`

Bash
```bash
authToken=`trinsic --json config | jq -r '.["auth token"]'`
```

PowerShell
```pwsh
$config = trinsic --json config | ConvertFrom-Json
echo $config.'auth token'
```

### DIFF
--- a/cli/src/macros.rs
+++ b/cli/src/macros.rs
@@ -6,7 +6,6 @@ macro_rules! dict {
     ($($key:expr => $value:expr,)+) => { dict!($($key => $value),+) };
     ($($key:expr => $value:expr),*) => {
         {
-            let _cap = dict!(@count $($key),*);
             let mut _map = std::collections::BTreeMap::new();
             $(
                 _map.insert($key, $value);
--- a/cli/src/macros.rs
+++ b/cli/src/macros.rs
@@ -1,4 +1,22 @@
 #[macro_export]
+macro_rules! dict {
+    (@single $($x:tt)*) => (());
+    (@count $($rest:expr),*) => (<[()]>::len(&[$(dict!(@single $rest)),*]));
+
+    ($($key:expr => $value:expr,)+) => { dict!($($key => $value),+) };
+    ($($key:expr => $value:expr),*) => {
+        {
+            let _cap = dict!(@count $($key),*);
+            let mut _map = std::collections::BTreeMap::new();
+            $(
+                _map.insert($key, $value);
+            )*
+            _map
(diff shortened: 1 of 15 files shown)


cli/src/macros.rs
```diff
@@ -1,3 +1,21 @@
+#[macro_export]
+macro_rules! dict {
+    (@single $($x:tt)*) => (());
+    (@count $($rest:expr),*) => (<[()]>::len(&[$(dict!(@single $rest)),*]));
+
+    ($($key:expr => $value:expr,)+) => { dict!($($key => $value),+) };
+    ($($key:expr => $value:expr),*) => {
+        {
+            let _cap = dict!(@count $($key),*);
```
geel9: Unused variable?

cli/src/services/mod.rs
```diff
@@ -36,5 +39,25 @@ pub(crate) enum Service<'a> {
...
-type Output = IndexMap<String, String>;
+type Output = BTreeMap<String, Item>;
+
+#[derive(Serialize, Debug, PartialEq)]
+#[serde(untagged)]
+pub(crate) enum Item {
+    String(String),
+    Json(Value),
+}
+//type Output = IndexMap<String, String>;
```
geel9: Should this commented block still exist (50-63)

geel9 (approved):

cli/src/macros.rs
tmarkovski: removed

cli/src/services/mod.rs
tmarkovski: removed

Issue #748, opened 2022-06-13
`Learn -> Platform` section


geel9:
Currently, the `Learn` docs section has two sections: `Key Concepts` and `Tutorials`

- Change `Key Concepts` to something like `SSI Concepts`, `Identity Concepts`, etc.
- Add `Platform` (or other name) section
  - Contains documentation around various aspects of the platform
  - `Webhooks`, `Login Flow`, `Authentication Tokens`, etc.
  - Start with just `Webhooks`

michaeldboyd:
@geel9 I'll be working on the docs this week, will assign this to myself

michaeldboyd:
- [ ] document authentication scheme
- [ ] document each individual event type 
- [ ] prep instruqt for webhooks and other how-to guides
- [ ] maybe add a verify webhook util function

michaeldboyd:
- [ ] document event types using protobuf objects in our references

PR #788, opened 2022-06-23
Split `Key Concepts` into `Key Concepts` and `Trinsic Platform`; rename `Services` to `Reference`

geel9:
This PR splits up the `Learn -> Key Concepts` section into two sections: `Key Concepts` and `Trinsic Platform`

- `Key Concepts` should introduce readers to the concepts behind *Identity/SSI* in general
- `Trinsic Platform` should document the intricacies of working with Trinsic *specifically*.
- This is just an organization change; the contents of these documents should likely change to better fit this categorization.

I can see some overlap between these two -- for instance, if we talk about credential schemas in a document in `Key Concepts`, we could link to the `Trinsic Platform -> Templates` document, IE, "here's how *we* handle this concept"

*Additionally*, this PR changes the `Services` tab to `Reference` -- it was named "services" to refer to the internal services that the SDK exposes; however, it communicated more of a business topic ("the services Trinsic offers") than what it actually was: an API reference.

Partially handles #748


tmarkovski (approved):
I agree with this breakdown

PR #795, opened 2022-06-27
Java multiple files option, `protoc-gen-sdk` options

fundthmcalculus:


geel9:
The proto changes should be submitted to the backend first, as the automation will overwrite the ones in this repo.

geel9 (approved):

fundthmcalculus:
> The proto changes should be submitted to the backend first, as the automation will overwrite the ones in this repo.

I'm well aware of that aspect 😁 , I'm trying to ensure that I'm not off in the weeds before I make the change in `server`.

sethjback:
This looks fine to me, with the caveat that I haven't used options like this enough to have an understanding of any potential unintended consequences.

Also, there was some discussion about just always including an oberon header if auth data was present - the backend will handle appropriately.

fundthmcalculus:
> This looks fine to me, with the caveat that I haven't used options like this enough to have an understanding of any potential unintended consequences.

Compilation is verified on my machine. That part isn't an issue.

> Also, there was some discussion about just always including an oberon header if auth data was present - the backend will handle appropriately.

I'm not opposed to this solution to the "inject metadata" problem, but in the event that we don't want to go the shotgun approach, this would enable it.

fundthmcalculus:
Closed as duplicate of https://github.com/trinsic-id/sdk/pull/797

PR #797, opened 2022-06-28
Copied Protos from server

fundthmcalculus:


proto/services/account/v1/account.proto
```diff
@@ -173,13 +173,20 @@ message AuthorizeWebhookResponse {
...
     // Login to account. If account doesn't exist, new will be created
-    rpc Login               (LoginRequest)              returns (LoginResponse);
+    rpc Login               (LoginRequest)              returns (LoginResponse) {
+        option (options.sdk_template_option).anonymous = true;
+    }
 
     // Confirm login step by responding to the challenge request
-    rpc LoginConfirm        (LoginConfirmRequest)       returns (LoginConfirmResponse);
+    rpc LoginConfirm        (LoginConfirmRequest)       returns (LoginConfirmResponse) {
+        option (options.sdk_template_option).ignore = true;
```
geel9, line 188: `LoginConfirm` should also be anonymous. (I'll issue a PR against server to fix)

geel9 (approved):

proto/services/account/v1/account.proto
geel9: On second thought, if `ignore` is `true`, no code is generated, so the `anonymous` flag doesn't matter.

fundthmcalculus: Correct. There is no reason to mark anonymous anything we don't generate.

PR #799, opened 2022-06-28
Copied Protos from server / Java 1 class update / Events.proto

fundthmcalculus:


fundthmcalculus:
Working on the Java build errors, this is intentional. The golang failures are independent and fixed here: https://github.com/trinsic-id/sdk/pull/801

geel9 (approved):

PR #825, opened 2022-06-30
Update Production Package Testing to 1.6.0

fundthmcalculus:


python/trinsic/__main__.py
```diff
@@ -1,5 +1,5 @@
 import asyncio
-import platform
+from asyncio import set_event_loop_policy
```
geel9:
```suggestion
from trinsic.trinsic_util import set_event_loop_policy
```

geel9 (changes requested):

fundthmcalculus:
node will continue to fail for reasons related to: https://github.com/trinsic-id/sdk/issues/838

geel9:
This PR breaks our docs by removing `sample_demo.py`, which is imported here:

https://docs-v2.trinsic.id/python/

Given that we have the python test script that can be executed via CLI, approval for me to commit to this branch removing this injection @fundthmcalculus?

fundthmcalculus:
Go for it. Thanks!

geel9 (approved):

Issue #822, opened 2022-06-30
SDK's should send metadata whenever they have it


fundthmcalculus:
We should send the metadata if the auth token profile exists, even if the endpoint doesn't require it. The server will simply ignore metadata that we don't need. This reduces complexity on the SDK side, as well as allows more auto generated template code.

Languages:
- [x] .NET #829 
- [ ] CLI
- [ ] Go
- [ ] Java
- [ ] Dart
- [ ] Python
- [ ] Ruby
- [ ] Swift
- [ ] Typescript

geel9:
Per a comment by @tmarkovski in #829 , it seems we may want to *not* always send an auth header.

We *do* need to always send metadata, so that we can add additional metadata (IE #845), but we want to control whether or not we send an auth header.

We have two options in my view:

- Always send auth header; force backend to deal with it.
- (My preference) Add an `authenticated: bool` arg to each lang's `BuildMetadata` function.
  - Defaults to `true`
  - Not a huge deal to automate in `protoc-gen-sdk`

geel9:
I've gone ahead with option #2 -- you can see the .NET implementation in #855 

If we decide to go another route this can be easily changed.

fundthmcalculus:
The solution to #845 automatically solves this problem. Closing this as a duplicate.

PR #829, opened 2022-06-30
.NET: Always send metadata; don't fail if auth token not present

geel9:
This PR changes .NET to always send message metadata in every call (even if unnecessary). Additionally, `BuildMetadata[Async]` no longer throws if `authToken` is not present; instead, it simply sends an empty auth header.

All in accordance with #822

### DIFF
--- a/dotnet/Trinsic/AccountService.cs
+++ b/dotnet/Trinsic/AccountService.cs
@@ -47,7 +47,7 @@ public class AccountService : ServiceBase
     /// <returns></returns>
     public async Task<string> SignInAsync(SignInRequest request) {
         if (string.IsNullOrWhiteSpace(request.EcosystemId)) request.EcosystemId = Options.DefaultEcosystem;
-        var response = await Client.SignInAsync(request);
+        var response = await Client.SignInAsync(request, await BuildMetadataAsync(request));
 
         var authToken = Base64Url.Encode(response.Profile.ToByteArray());
 
@@ -65,7 +65,7 @@ public class AccountService : ServiceBase
     /// <returns></returns>
     public string SignIn(SignInRequest request) {
         if (string.IsNullOrWhiteSpace(request.EcosystemId)) request.EcosystemId = Options.DefaultEcosystem;
-        var response = Client.SignIn(request);
+        var response = Client.SignIn(request, BuildMetadata(request));
 
         var authToken = Base64Url.Encode(response.Profile.ToByteArray());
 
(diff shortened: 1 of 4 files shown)


fundthmcalculus (commented):
@sethjback any changes needed on the server side for this?

sethjback:
No - the server looks for metadata if it is required, otherwise it is ignored.

sethjback (approved):

fundthmcalculus (commented):
In light of wanting to send version information in the header, I think that we should probably always send the metadata `Authorization` as empty.

geel9:
> In light of wanting to send version information in the header, I think that we should probably always send the metadata `Authorization` as empty.

I changed it to not send the header at all because, upon reviewing the backend codebase, functionality depends on the presence of the header whatsoever -- not whether it's empty or not.

Although it depends on whether or not an *empty* header is serialized and sent down the wire? Likely language-specific?

fundthmcalculus:
> > In light of wanting to send version information in the header, I think that we should probably always send the metadata `Authorization` as empty.
> 
> I changed it to not send the header at all because, upon reviewing the backend codebase, functionality depends on the presence of the header whatsoever -- not whether it's empty or not.
> 
> Although it depends on whether or not an _empty_ header is serialized and sent down the wire? Likely language-specific?

Good points. @sethjback thoughts on backend behavior? I agree, we shouldn't rely on language-specific behavior here. I think if the metadata is required, it checks it. If it isn't required, it ignores it (automatically becomes effectively anonymous).

sethjback:
If we add the sdk / okapi version as metadata we will always want to send those - so the way it's implemented here makes sense: always call the build metadata. Set the sdk/okapi versions, and if appropriate `authorization`.

As for an empty `authorization` header - it would be easy to add a `NoResult()` return if the header is empty as well as missing. If that would save time/effort on all the SDKs that is probably the best route to go.

geel9:
Seems good to me -- I've created a PR against `server` to handle empty Auth headers @sethjback

tmarkovski:
The unintended side effect of this is that if, for some reason, the `authToken` is present, but invalid/expired/different_eco, the user won't be able to login, because the server will always try to check the authorization if the header is present.
In general, `Authorization` should be sent if the server endpoint is decorated with `Authorize`. If it's `AllowAnonymous` it's best not to send `Authorization` to prevent side effects.

I'm sorry, I missed the convo on this.

sethjback:
@tmarkovski would it affect login? If the authorization header is internally invalid it would, but if it checks out from a signature/timestamp perspective login should succeed since anonymous endpoints don't do additional authz.

It almost seems better as well to fail if there is an internally inconsistent auth header set, even on login, to give feedback to the sdk user.

It is not strictly correct to send an `authorization` header to anonymous endpoints. That being said, the server should handle the situation correctly.

Maybe I'm not fully appreciating the scenario here though...

PR #855, opened 2022-07-08
.NET Metadata: SDK Version/Lang; Okapi Version; Optional Authentication

geel9:
- Adds `authenticated` bool param to `BuildMetadata`, `BuildMetadataAsync`
  - If `false`, authentication header will not be constructed for request -- suitable for anonymous calls
  - If `true`, authentication header will be constructed for request, **and exception will be thrown** if no auth token exists
    - Behavior is the same as before we switched to always sending metadata
- Sends `TrinsicSDKVersion`, `TrinsicSDKLanguage`, `TrinsicOkapiVersion` headers with all requests

**Unresolved issues**
- The `TrinsicSDKVersion` header will always be set to `1.0.0.0` as I don't quite understand how our library's version is actually being set when we publish it; nor how to fetch this at runtime.

**Further work**
- I must update the `protoc-gen-sdk` template to factor in the `anonymous` option when calling `BuildMetadata`


fundthmcalculus (approved):
We discussed, and I fixed the bits that needed it. @geel9 if you could review the changes I made and ensure you agree on implementation, that would be great.

dotnet/Trinsic/ServiceBase.cs
```diff
@@ -72,30 +72,32 @@ public abstract class ServiceBase
...
     protected GrpcChannel Channel { get; }
 
+    private MetadataResponse? _okapiMetadata;
     /// <summary>
     /// The cached metadata of the Okapi library being used.
-    ///
-    /// Do not read this variable directly -- use `GetOkapiMetadata()`.
     /// </summary>
-    private MetadataResponse? CachedOkapiMetadata { get; set; } = null;
+    private MetadataResponse CachedOkapiMetadata {
```
geel9, line 79: Good catch

PR #860, opened 2022-07-09
Dart: metadata request, version signature

fundthmcalculus:


geel9 (changes requested):
Metadata should be built for every request (incl. `signIn`, `login`, `loginConfirm`, `createEcosystem`)

geel9 (approved):

PR #861, opened 2022-07-09
Go metadata sending

fundthmcalculus:


geel9 (changes requested):
Metadata should be built for every request (incl. `SignIn`, `Login`, `LoginConfirm`, `CreateEcosystem`)

fundthmcalculus:
> Metadata should be built for every request (incl. `SignIn`, `Login`, `LoginConfirm`, `CreateEcosystem`)

Can do, that also entails identifying which calls are actually anonymous, something I will do in a separate PR because it affects the protos.

geel9 (approved):

PR #862, opened 2022-07-11
Java updates with metadata options

fundthmcalculus:


geel9 (approved):

geel9 (changes requested):
Metadata should be built for every request (incl. `Login`, `LoginConfirm`, `CreateEcosystem`)

geel9 (approved):

PR #863, opened 2022-07-11
Special handling of `map<>` proto types in documentation

### LINKED ISSUE #837: Better documentation support for map fields
Our protobuf documentation generation currently handles `map` fields in the following way:

This is due to how `protoc` handles maps: it creates an `Entry` type with `key` and `value` fields, and then sets the original map field to just be an array of `Entry`s.

Our documentation generation should handle this special case and display the types in a more natural form -- abstracting away the interstitial `Entry` types.

geel9:
### Overview

Given the following protobuf message:

```protobuf
message Foo {
    map<string, bool> bar = 1;
}
```

The protobuf compiler will produce the following interstitial types during compilation:

```protobuf
message Foo {
   repeated BarEntry = 1;
}

message BarEntry {
   string key = 1;
   bool value = 2;
}
```

This causes our protobuf documentation tooling to generate documentation according to the second structure -- confusing developers, as **this is not the actual structure generated for most programming languages**.

### Solution

First, `protoc-gen-json` was modified to export an `is_map_entry` bool on Messages which are interstitial `Entry` types.

Then, our protobuf documentation tooling handles this special case uniquely:

- Changes the type name to be `map(key_type -> value_type)`
- Short-circuits the expandable subobject to expand into the `value_type` object, as opposed to expanding into the interstitial `Entry` type.
  - This also correctly handles situations where `value_type` is primitive (so should not be expanded) or is an enum.
- Adds array iterator indicator to field context
  - For simple `repeated` types, adds an `[i]` after the variable name to preserve array context
  - For map types, this is `[key]`

Closes #837

### DIFF
--- a/docs/_static/extra.css
+++ b/docs/_static/extra.css
@@ -165,6 +165,10 @@ input {
     color: grey;
 }
 
+.proto-context-array-key {
+    color: #959fa7;
+}
+
 
 
 /** 
--- a/docs/_static/proto.json
+++ b/docs/_static/proto.json
@@ -4533,6 +4533,7 @@
       "name": "FieldsEntry",
       "full_name": "services.verifiablecredentials.templates.v1.CreateCredentialTemplateRequest.FieldsEntry",
       "description": "",
+      "is_map_entry": true,
       "fields": [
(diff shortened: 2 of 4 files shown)


fundthmcalculus (approved):
NICE! @michaeldboyd take a look at this!

PR #867, opened 2022-07-11
Remove `default_ecosystem` from CLI config

tmarkovski:


geel9:
I think this parameter might have functionality and value that it didn't have in the SDKs.

At the very least, it makes our tutorials cleaner.

fundthmcalculus (approved):

tmarkovski:
> I think this parameter might have functionality and value that it didn't have in the SDKs.

Can you expand on this please?

geel9:
@tmarkovski We removed the `default_ecosystem` value from the SDKs because it seemed redundant -- it's a config value that is set at runtime and which sometimes, but not other times, took effect.


With the CLI, the use case is different: the config value is saved to disk, and it pretty much only applies to the `login` calls, which I can foresee being used a lot for testing / prototyping / general dev work by our customers. A user of the CLI can just set the config value and leave it there for as long as they like.


Having to explicitly specify `--ecosystem` whenever you login is going to get annoying, *especially* if you are using an autogenerated ecosystem with a long and difficult to remember name.

tmarkovski:
I'd like to challenge this to make sure I understand this will be a practical problem.
You're implying this would be annoying when typing it manually; in automated scenarios it's less annoying.

If you're going to work with auto-generated ecosystem, you're most likely doing some testing work. How many times are you going to login to that system on the same machine? Is it more annoying to say `login --ecosystem <eco>` or `config --default_eco <eco> && login` ? You're going to have to set the ecosystem at least once, regardless of which path it's taken.

Do we see users having to login with the CLI manually multiple times for multiple users? Is that a likely scenario (outside our internal testing scenarios)?

PR #877, opened 2022-07-14
anonymize methods that can be anonymized, rename `searchWallet`

### LINKED ISSUE #864: Verify that all anonymous methods are properly marked in the proto files
This should be a `.proto` change on the server side, and will then be reflected here.

Languages:
- [ ] CLI
- [ ] Docs
- [ ] .NET
- [ ] Dart
- [ ] Go
- [ ] Java
- [ ] Python
- [ ] Ruby
- [ ] Swift
- [ ] Typescript

fundthmcalculus:
closes #864


devops/generate_proto_files.py
```diff
@@ -155,18 +155,21 @@ def run_protoc(
...
-        arg_list if isinstance(arg_list, list) else [arg_list]
-        for arg_list in command_args
-    ]
-    command_args = list(itertools.chain(*command_args))
-    # Strip blank arguments because protoc WILL DIE, and do so passive-aggressive
-    command_args = [arg for arg in command_args if arg]
-    logging.info(command_args)
-    if os.system(" ".join(command_args)) != 0:
-        raise Exception("protoc failed")
+    # Because of the length of some command line arguments, we have to compile each file individually
```
geel9: This breaks `protoc-gen-json` as it needs to be run against all the proto files at once.

What can we do about this?

fundthmcalculus: Let me think about options on this. In general, a plugin shouldn't care?

geel9: That's because plugins typically generate one file per input `.proto` file, but `protoc-gen-json` generates a single output JSON file combining information from all the input files.

It wouldn't be too much work to write a script which would combine all the JSON files into a single file, and run that in post. I'd prefer not to do that, though.

Is the issue that the command is literally just too large when specifying all proto files?

fundthmcalculus: It can be as a path error. Windows has a command-line limitation. What we could do to get around it is relative path all the things, I think. Hang on. For now, I'll just disable this change.

geel9 (approved):

PR #890, opened 2022-07-15
Automatically host preview version of docs for PRs

geel9:
This PR creates a new GitHub action which runs on every PR that touches docs; it builds that version of the docs and hosts a static site in netlify.

### DIFF
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -56,7 +56,7 @@ jobs:
           NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }} 
         run: |
           echo "Creating site ${{ env.DOCS_BRANCH_NAME }}"
-          netlify sites:create --name "${{ env.DOCS_BRANCH_NAME }}"-a josh-abi1wqk
+          netlify sites:create --name "${{ env.DOCS_BRANCH_NAME }}" -a josh-abi1wqk
       - name: Wait for Netlify site to be ready
         continue-on-error: true
         working-directory: ./docs-site/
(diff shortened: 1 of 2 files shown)


.github/workflows/test-docs.yml
```diff
@@ -0,0 +1,62 @@
+name: Documentation (PR preview)
+
+on:
+  pull_request:
+    paths:
+      - "docs/**"
```
fundthmcalculus: Docs/**/*

fundthmcalculus (changes requested):
See comment. Also, requesting changes to block merge. Will approve when done.

janpieterz:
Interesting. Do we use Netlify for our normal docs?

geel9:
> Interesting. Do we use Netlify for our normal docs?

No -- we use GitHub Pages.

The normal docs build process commits to the `gh-pages` branch of this repo, which is served from our docs site.

However, there is no way to have more than one branch of GitHub pages -- we'd need to either make a separate repository, or do some other kind of hackery.

Initially I wanted to use Azure Storage for this, which works fine when we're just deploying a single static site, but deploying multiple static sites is... tricky. 

That is -- with Azure, if we deployed a single static site to the root directory of a storage container, it works just fine, but to support multiple PRs having previews available concurrently, we'd have to place each branch into a subdirectory of said container, which breaks when relative URLs are at play on the docs site (which is in a lot of places)


The Azure solution would be to create a new storage unit for each branch, configure it appropriately, generate an access token for it, and then deploy the site to it.

I'm A) not familiar enough with Azure and B) not confident that this is a sane solution.



So, I searched for a service which is designed to programmatically host static sites, and Netlify seems like it's going to be a good fit. I think we'll even be able to use their free tier indefinitely for this purpose.

.github/workflows/test-docs.yml
geel9: @fundthmcalculus Ready for re-review

fundthmcalculus (approved):
Nice!! @janpieterz

PR #896, opened 2022-07-26
[cli] add library target

### LINKED ISSUE #790: Refactor CLI into a CLI wrapper and Rust SDK
Per a discussion with @geel9 rust doesn't have an SDK, but the CLI does effectively that. We should refactor these two applications to enable the use of our v2 services from Rust. Separation of concerns here.

- [ ] Add `rust` to the "SDK Change Request" issue template
- [ ] Refactor rust code into a crate
- [ ] Crate publish
- [ ] Make CLI wrapper consume Rust SDK

~Opening a can of worms here, but should we consider moving the CLI into a language other than Rust.~ **Update** we won't do this due to discussion below. @geel9 and I talked about this 2022-06-23, and there are some benefits to doing so:

1. Rust is a slower language to write, less flexible.
2. @tmarkovski is the only strong rust programmer on the team (creates a bus factor problem) / CLI is something we should all be able to support
3. Rust does compile to native binaries, eliminating CLI 3rd party dependencies (.NET, Python, Java)
4. Rust is faster, but do we care for the CLI?
5. .NET is broadly available, and more well known. Caveat is installing .NET core to use it.
6. Golang is native binary, but the lack of support for Windows is major issue
7. Use Rust, but refactor out a Rust SDK separately (see original comments)
8. Other options?

tmarkovski:
Adds lib artifacts so that the package can be used and referenced as library

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -31,6 +31,7 @@ tonic-build = { version = "0.6", features = ["prost", "rustfmt"] }
 
 [lib]
 crate-type = ["cdylib", "rlib"]
+doctest = false
 
 [[bin]]
 name = "trinsic"
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -5,6 +5,7 @@ authors = ["Trinsic <support@trinsic.id>"]
 edition = "2021"
 license = "Apache-2.0"
 description = "Command-line tool for Trinsic APIs"
+default-run = "trinsic"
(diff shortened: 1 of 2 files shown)


fundthmcalculus (approved):
Does this close the issue #790 ?

tmarkovski:
> Does this close the issue #790 ?

It does, in my opinion. I've linked it to this PR

PR #899, opened 2022-07-27
Sdk 895 - Split Node/web packages

### LINKED ISSUE #895: Split SDK typescript package for node/web
Due to bundlers other than webpack not honoring the `"browser"` tag in `package.json`, we need to split the node and web typescript packages for our SDK similar to how we split the okapi package. Ideally, we can keep a common codebase, and just rename the `package.json` on publish (which is what we do for okapi). There would be some idiosyncracies with the different package dependencies at the SDK level, so we might need to make two different `package.json` files (if that's possible?)

fundthmcalculus:
I have this working with subpackage locations and provider setting. I dislike how the tests are duplicated, so I'm open to thoughts on how to avoid that. I do need to mimic importing from `./index` so that we can set the node/web trinsic provider.

@MichaelEdwardBlack thoughts?


MichaelEdwardBlack:
We discussed over gather but posting here for visibility:

Moved common tests outside the packages and left the tests requiring the environment-specific provider inside their respective packages. If there was a way to change which package/provider was being used based on whether jest is running or karma is running then that would be awesome! But, I don't think there's a way to do that so what you've done is looking really good! :)

fundthmcalculus:
Node 14 fails for weird reasons, will need to investigate

fundthmcalculus:
Decided against this method - closing.

PR #901, opened 2022-07-28
[ts][go] Add memory token provider for storing auth tokens

fundthmcalculus:
This is better done before the node/web SDK package changes, so here's the PR for go and typescript. Not a blocker for release of 1.7.0, but nice to have.


tmarkovski (approved):

go/services/token_provider.go
```diff
@@ -0,0 +1,37 @@
...
+		return val, nil
+	}
+	return "", nil
+}
+
+func (m *memoryTokenProvider) Save(authToken string, name string) {
+	m.tokens[name] = authToken
+}
+
+func NewMemoryTokenProvider() TokenProvider {
```
tmarkovski, line 33: Should this be just `New()` or is full name idiomatic to go?

fundthmcalculus: More idiomatic to say `New[Classname]` since go doesn't really have classes or constructors (just struct definitions).

If (when) we have `FileTokenProvider` or something, it would be good to make them unambiguous.

PR #906, opened 2022-08-01
Export `@trinsic/trinsic/browser` properly

fundthmcalculus:


tmarkovski (commented):
Shouldn't this load from lib or are we keeping everything in src now?

fundthmcalculus:
I tried that. That was what required the lib browser import directive. Apparently you can't have a default import from a lib folder from what I can tell.

fundthmcalculus:
This looks promising: https://github.com/nodejs/node/issues/14970#issuecomment-570036324

fundthmcalculus:
Appears to work locally, testing with one final rc before sending https://github.com/trinsic-id/sdk/issues/881

fundthmcalculus:
It works as expected.

tmarkovski (approved):
\o/

geel9 (approved):

PR #911, opened 2022-08-02
Let's get the right export directives, browser friendly

fundthmcalculus:


fundthmcalculus:
React apparently doesn't like this, but certain other typescript systems do, so what gives? Ideally, I can compile `node.ts` and `browser.ts` in place. Hmmmm

tmarkovski (approved):

PR #1001, opened 2022-08-17
Update migrate.md

rileyphughes:
What is the value of the "who should consider migration" section? shouldn't it be everyone?

On wallets, it's important we get messaging right. We should call our old wallets "custodian wallets" and new wallets "hybrid wallets" IMO

I added a "without the same privacy guarantees" to the revocation section to make sure it's clear the trade-off we made going with status list


tmarkovski (approved):
LGTM

fundthmcalculus:
> [Preview docs site for rileyphughes-suggested-edits@9d2df075c66e1b0a9427a735b40be155d7ab26d6](https://trinsic-docs-pr-9d2df075c66e1b0a9427a735b40be155d7ab26d6.netlify.app/)

This is slick, thanks Josh! Also, collaborating on docs this way seems cleaner than some of the other tools we've used. Markdown is so nice. :)

PR #1003, opened 2022-08-18
Return a wrapper that contains the base64 auth token

### LINKED ISSUE #1002: `CreateEcosystem()` returns byte[] instead of base64 profile string

fundthmcalculus:
Closes trinsic-id/server#1278


fundthmcalculus:
Investigating the node-14 situation currently, will fix as part of this PR - it has to do with `AbortController`

geel9:
I feel like this should be a backend change so that we can reproduce it in all languages with minimal effort.

I definitely think this shouldn't be relegated to just TypeScript.

fundthmcalculus:
> I feel like this should be a backend change so that we can reproduce it in all languages with minimal effort.
> 
> I definitely think this shouldn't be relegated to just TypeScript.

Good call. Closing this PR.

PR #1009, opened 2022-08-21
Configure TrinsicService with DI pipeline. Docs update.

tmarkovski:
- Added the TrinsicService to the DI extensions
- Introduced builder pattern to configure service options and token persistence
- Updated documentation for using the dotnet sdk in common scenarios
  - Removed documentation block for Configuration which causes confusion about using the SDK in server scenario. Code block applied to client-side


fundthmcalculus (approved):

sethjback (approved):
:tada:

PR #1012, opened 2022-08-26
Update build-proto.yml to support `protoc-gen-sdk` automatically

fundthmcalculus:
Need to figure out how to handle the submoduled `swift` side of the problem (similar action in that repository?), but let's figure out the other languages first.

- [x] Should we remove `debug-service` completely, or properly expose & support it? It's generated already, but buried in the lower level code.


fundthmcalculus:
A lot of the failing tests here will be fixed by:
https://github.com/trinsic-id/sdk/pull/1022

MichaelEdwardBlack (approved):

PR #1022, opened 2022-08-31
Fix server copied protos

fundthmcalculus:


fundthmcalculus:
switching to dev endpoint while waiting on staging bugfix.

MichaelEdwardBlack (approved):

PR #1051, opened 2022-09-07
Updated server copied protos

fundthmcalculus:
* import directive fixes from #1049  (here to prevent server `git push force` overwrites)
* also `prettier` fix on the entire directory.


janpieterz:
Don't think this is reviewable, we should just merge them imho

web/src/ProviderService.ts
```diff
@@ -23,10 +23,11 @@ import {
...
     InviteResponse,
-    ProviderDefinition, RefreshVerificationStatusRequest,
-    RefreshVerificationStatusResponse,
-    RetrieveVerificationRecordRequest,
-    RetrieveVerificationRecordResponse,
+    ProviderDefinition,
+    RefreshDomainVerificationStatusRequest,
+    RefreshDomainVerificationStatusResponse,
+    RetrieveDomainVerificationRecordRequest,
+    RetrieveDomainVerificationRecordResponse,
```
fundthmcalculus, line 30: This is the only real change

janpieterz (approved):

tmarkovski (approved):

PR #1053, opened 2022-09-08
Copied Protos from server

fundthmcalculus:


fundthmcalculus:
Merging auto-generated code post-import fix.

PR #1056, opened 2022-09-08
Update build-proto.yml

fundthmcalculus:
* Prevent `git diff` from spamming `stdout` when all we want to do is figure out if *any changes exist*


fundthmcalculus:
confirmed that logs for relevant step no longer contain the diff/patch data.

PR #1059, opened 2022-09-13
Server 843 consume public ecosystem styling

fundthmcalculus:
Relates to https://github.com/trinsic-id/server/pull/941

* Also converted demo to typescript.
* `setupDemo.ts` is used to set up a sample ecosystem with a domain, and CDN images.

image


janpieterz:
- That logo doesn't seem styled right (but is different PR?)
- Why is that bar of "Select a credential" so green (but is different PR?)

Besides that this is just a sample to test, so if it works I'm fine with it for now.

janpieterz (approved):

fundthmcalculus:
> That logo doesn't seem styled right (but is different PR?)

That's a different issue since I use an icon of the wrong shape to demonstrate icon being pulled from our CDN.

> Why is that bar of "Select a credential" so green (but is different PR?)

That's the color I selected for the "light" display configuration, just showing it to the user in the easiest way possible. #notadesigner

> Besides that this is just a sample to test, so if it works I'm fine with it for now.

It does work, hence the screenshot - thanks for checking on this!

PR #1073, opened 2022-09-20
Server copied protos import fix

fundthmcalculus:
* Typescript is a known protoc-gen-sdk issue of updating the imports
* Golang was a miss in the python line replace.


geel9 (approved):
Thank you!

PR #1077, opened 2022-09-20
DID Upgrade Documentation

geel9:
Adds documentation and sample code for DID Upgrades

### DIFF
--- a/docs/reference/services/provider-service.md
+++ b/docs/reference/services/provider-service.md
@@ -270,6 +270,70 @@ Deletes a webhook from an ecosystem.
 
 {{ proto_method_tabs("services.provider.v1.Provider.DeleteWebhook") }}
 
+---
+
+## Upgrade Wallet DID
+
+!!! warning "Restricted API"
+    This feature is not yet publicly available, and the list of supported DID Methods is not final.
+
+    Please contact Trinsic to enable this endpoint on your account.
+
+!!! info "Ledger Interactions"
+    Many DID methods are based on distributed ledgers (commonly known as "blockchains"), so the following should be kept in mind when registering a DID which uses a ledger:
+
+    - Depending on the DID method chosen, registering a DID may incur a ledger write fee, which is different for each ledger. **You are responsible for these fees.**
+        - We will provide clear documentation regarding these fees before this feature becomes generally available.
+    - Newly-registered DIDs may not be immediately resolvable, as most ledgers cannot provide instantaneous writes.
+        - Issuance and verification of credentials against these DIDs will still function during this time, but only within the Trinsic platform.
+
+Upgrades a wallet's DID from the default `did:key` to another DID Method. This endpoint may only be called by an ecosystem provider.
+
+Trinsic will register a DID Document for you, and update the wallet's `public_did` property to the newly-registered DID.
+
+{{ proto_sample_start() }}
+    === "Trinsic CLI"
+        ```bash
+        trinsic provider upgrade-did --wallet-id {wallet_id} --method ion --method-options testnet
+        ```
+
+    === "TypeScript"
+        <!--codeinclude--> 
+        ```typescript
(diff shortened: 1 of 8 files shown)


docs/reference/services/provider-service.md
```diff
@@ -270,6 +270,70 @@ Deletes a webhook from an ecosystem.
...
+    Many DID methods are based on distributed ledgers (commonly known as "blockchains"), so the following should be kept in mind when registering a DID which uses a ledger:
+
+    - Depending on the DID method chosen, registering a DID may incur a ledger write fee, which is different for each ledger. **You are responsible for these fees.**
+        - We will provide clear documentation regarding these fees before this feature becomes generally available.
+    - Newly-registered DIDs may not be immediately resolvable, as most ledgers cannot provide instantaneous writes.
+        - Issuance and verification of credentials against these DIDs will still function during this time, but only within the Trinsic platform.
+
+Upgrades a wallet's DID from the default `did:key` to another DID Method. This endpoint may only be called by an ecosystem provider.
+
+Trinsic will register a DID Document for you, and update the wallet's `public_did` property to the newly-registered DID.
```
janpieterz: Could you maybe write a sentence or two about:
1. It not being possible to reverse this, once done it's out there
1. The impact (e.g. what about already issued credentials from this wallet)
1. The benefits?

janpieterz (approved):
lgtm, little nitpicks re adding some context to the docs but just throw something on there that feels good to you ;)

dotnet/Tests/Tests.cs
```diff
@@ -26,12 +26,12 @@ namespace Tests;
...
 public class Tests
 {
-#if DEBUG
+#if DEBUG && false
     private const string DefaultEndpoint = "localhost";
     private const int DefaultPort = 5000;
     private const bool DefaultUseTls = false;
 #else
-    private const string DefaultEndpoint = "staging-internal.trinsic.cloud";
+    private const string DefaultEndpoint = "dev-internal.trinsic.cloud";
```
tmarkovski: We test SDK against staging. Forgot to revert this or is there another reason?

dotnet/Tests/Tests.cs
```diff
@@ -26,12 +26,12 @@ namespace Tests;
 [SuppressMessage("ReSharper", "MethodHasAsyncOverload")]
 public class Tests
 {
-#if DEBUG
+#if DEBUG && false
```
tmarkovski: Intentional disable?

fundthmcalculus: This prevents localhost ever being used, is that intentional? Also, default should be staging, not dev for testing purposes.

dotnet/Tests/Tests.cs
fundthmcalculus:
```suggestion
    private const string DefaultEndpoint = "staging-internal.trinsic.cloud";
```

dotnet/Tests/Tests.cs
```diff
@@ -26,12 +26,12 @@ namespace Tests;
...
     private const string DefaultEndpoint = "localhost";
     private const int DefaultPort = 5000;
     private const bool DefaultUseTls = false;
```
fundthmcalculus:
```suggestion
    private const string DefaultEndpoint = "dev-internal.trinsic.cloud";
    private const int DefaultPort = 443;
    private const bool DefaultUseTls = true;
```

dotnet/Tests/Tests.cs
fundthmcalculus:
```suggestion
#if DEBUG
```

dotnet/Tests/Tests.cs
fundthmcalculus: Beat me to it! 😁

geel9: Accidental inclusion -- thanks for catching this

PR #1078, opened 2022-09-22
Show Docs site with links directly to changed files

fundthmcalculus:


devops/build_sdks.py
```diff
@@ -248,6 +286,10 @@ def build_none(args) -> None:
 def parse_arguments():
     parser = argparse.ArgumentParser(description="Process SDK building")
     parser.add_argument("--package-version", help="Manual override package version")
+    # TODO - Get the SHA and head-ref programmatically?
```
geel9: Could be done via environment variables -- `GITHUB_HEAD_REF`, `GITHUB_SHA`, `DOCS_BRANCH_NAME`

devops/build_sdks.py
```diff
@@ -267,6 +309,7 @@ def main():
         "java": build_java,
         "dart": build_dart,
         "typescript": build_typescript,
+        "doc_site": build_docs_site,
```
geel9: Could we change this to something more explicitly "github-actions-automation"y?

IE, `docs_pr_automation`?

geel9 (approved):
Incredible!

devops/build_sdks.py
fundthmcalculus: True, I'm trying to keep the interface cleaner than that, per-se. I don't like passing information on both command line and env vars. If I had to use the github token, I'd definitely run that through the env var.

PR #1097, opened 2022-10-05
Create ecosystem doesn't set the token as current

tmarkovski:


janpieterz (approved):
Lgtm

fundthmcalculus:
Do we need to do this for all languages?

PR #1098, opened 2022-10-06
Copied Protos from server

mewmba:


fundthmcalculus:
Will investigate failures, this is partially a performance issue with overloading the server api and cosmos db.

tmarkovski:
> Will investigate failures, this is partially a performance issue with overloading the server api and cosmos db.

What do you mean overloading the api and db? What action overloaded them? We should be able to handle a "CI load".

fundthmcalculus:
> > Will investigate failures, this is partially a performance issue with overloading the server api and cosmos db.
> 
> What do you mean overloading the api and db? What action overloaded them? We should be able to handle a "CI load".

Cosmos db errors about rate limit exceeded. We triggered all tests on every language, every version.

PR #1102, opened 2022-10-10
Sdk 1101 docs authorship fix

### LINKED ISSUE #1101: Docs authorship is incorrect
I am not the sole writer of the [Migration Guidelines](https://docs-v2.trinsic.id/learn/migrate) (I only fixed a single typo) -- yet I am the only author listed.

Perhaps it's not fetching deep enough?

fundthmcalculus:


docs/overrides/main.html
```diff
@@ -1,37 +1,49 @@
 {% extends "base.html" %}
 
+{% block content %}
+  {{ super() }}
+
+  {% if git_page_authors %}
```
geel9, line 6: This causes a second authorship section to appear

geel9 (changes requested):

docs/overrides/main.html
fundthmcalculus: weird, because it doesn't locally. :/

geel9 (approved):
Thank you!

PR #1110, opened 2022-10-14
Go remove okapi

sethjback:
PR Removes need for `okapi` and instead uses native oberon library.

Benefits:
1. Removes CGO and all its headaches
2. Simplifies setup of the services SDK (doesn't require shipping native binaries)


geel9:
I think the benefits are compelling -- the largest drawback I can see is that this brings Go out of step with every other SDK we have where we _do_ use Okapi. 

Additionally, if we ever need to do anything more with Okapi on the SDK side (say, local verification), then we're right back to square one.

That said, I think developer usability is king, and having to install a separate set of binaries and use CGO seriously impacts that -- so I'm in favor.

fundthmcalculus:
> I think the benefits are compelling -- the largest drawback I can see is that this brings Go out of step with every other SDK we have where we _do_ use Okapi.

True, but this issue isn't bad. The bigger goal is to keep the API surface consistent among all the SDKs, not keep the underlying implementation details consistent.

> 
> Additionally, if we ever need to do anything more with Okapi on the SDK side (say, local verification), then we're right back to square one.

Yes and no. Directly pulling in the CGO binaries is easier than pulling them in 2 levels up the chain. Also, if we can keep our network SDK and our "offline" SDK more separate.

> That said, I think developer usability is king, and having to install a separate set of binaries and use CGO seriously impacts that -- so I'm in favor.

Totally agree with making developer usability better. That said, golang community's attitude of "just rewrite it in go" is equally annoying, and certainly doesn't help their case.


@sethjback as we discussed, typescript might benefit from this as well, if we can remove the 2MB WASM binary that is required for anything.

PR #1115, opened 2022-10-18
Enable JDK 11 support, also update to relevant okapi 1.6.1

### LINKED ISSUE #1114: Enable OpenJDK 11 compatibility
OpenJDK 11 is still used, and is in support. We need to change the class file version to `55.0` or earlier. Ideally, this is just a `build.gradle` configuration change.

fundthmcalculus:
closes #1114


fundthmcalculus:
Tested locally confirmed working. Build failures due to something else.

PR #1138, opened 2022-11-01
Replace failing fontawesome icon with material

lanceATtrinsic:
font awesome icon from javascript wasn't resolving... changed to use material icons like the other links


lanceATtrinsic:
Tested as working correctly

BEFORE
Screen Shot 2022-11-01 at 1 05 10 PM

AFTER
Screen Shot 2022-11-01 at 1 05 22 PM

geel9 (approved):

PR #1148, opened 2022-11-04
Update index.md

rileyphughes:
fixed too many line breaks

but also I want my name to show up in the authors, since the second paragraph is coming in the first-person from me


fundthmcalculus:
Authorship is auto calculated, so it should update when we deploy. You could also sign it, since that section is more of a letter from you to our customers.

fundthmcalculus (approved):

PR #1169, opened 2022-11-17
mod to vaccination.md

srehorn:
Just a little clarification about where to find the value to use for the TEMPLATE_ID (DX nitpicking :)


fundthmcalculus (approved):
Thank you for the good catch! :beers:

PR #1178, opened 2022-11-25
Make SDK Actions Run Again

fundthmcalculus:
Expose `AccessManagementService` typescript
Fix Java test error
Update action download artifact to allow failed runs (okapi with iOS target failure).


cli/src/parser/vc.rs
```diff
@@ -54,6 +54,7 @@ fn create_proof(args: &ArgMatches) -> Result<Command, Error> {
         reveal_document: args.value_of("reveal-document-file").map(|x| x.into()),
         document_file: args.value_of("document-file").map(|x| x.into()),
         item_id: args.value_of("item-id").map(|x| x.into()),
+        nonce: args.value_of("nonce").map(|x| x.into()),
```
fundthmcalculus: @tmarkovski thoughts on how this is implemented? I need to update the cli powershell script, but want to confirm this implementation before I merge.

tmarkovski: This is fine. You can also simply pass `nonce: vec![],` in the `create_proof` service, just to make the compiler happy, and not have to implement the actual flag in the CLI, since we need to figure out how we want to pass that from the command line and convert it to byte array.

cli/src/proto/sdk/options/v1/mod.rs
```diff
@@ -1,19 +1,20 @@
 /// Configuration for Trinsic SDK Services
-#[derive(::serde::Serialize, ::serde::Deserialize, Clone, PartialEq, ::prost::Message)]
+#[derive(::serde::Serialize, ::serde::Deserialize)]
+#[derive(Clone, PartialEq, ::prost::Message)]
 pub struct ServiceOptions {
     /// Trinsic API endpoint. Defaults to `prod.trinsic.cloud`
-    #[prost(string, tag = "1")]
```
tmarkovski: Need to run `cargo fmt` here

cli/src/parser/vc.rs
fundthmcalculus: Will do that. I wasn't sure how to best pass the nonce in, and for now we can assume default empty.

tmarkovski (approved):

PR #1179, opened 2022-11-28
tests: Add node tests for CredentialService and Templates

janpieterz:


fundthmcalculus (approved):
I thought we already had this test in the VaccineShared.ts file?

janpieterz:
@fundthmcalculus Mmmh, yea, maybe? it's tough to compare those tests, I just noticed we ran some on Browser but not node during a debug session so tackled this. We can refactor this a bit in the future.

fundthmcalculus:
The shared component is there so we aren't duplicating code between node and browser. It looks like all the checks failed, so something is broken.

janpieterz:
I'll check it out :)

PR #1183, opened 2022-12-02
add dotnet interceptor options

fundthmcalculus:


tmarkovski:
Is there an issue related to this PR?

fundthmcalculus:
Yes, in the server repo.

janpieterz (changes requested):
I'd drop this whole PR. Though interesting, we don't use it for tracking traffic origins for now, so let's keep the SDK API clean (the PR will survive)

PR #1193, opened 2022-12-06
Copied Protos from server

mewmba:
:crown: copied proto files from server repo


tmarkovski (approved):

tmarkovski:
How do we fix this issue with PR validation always failing in a way that has low maintenance? Can the generator simply run the proto generation script and then update it using the language specific formatting? Can this be done without using the templating engine?

There's also issues with spaces, tabs, comments etc, that always produce changes. Are we happy with the maintenance level the protoc gen plugin?

fundthmcalculus:
> How do we fix this issue with PR validation always failing in a way that has low maintenance? 

Our SDK actions have always been slightly flaky when slamming the backend with so many concurrent requests. Golang in this case is failing due to the enforcement of `"default"` ecosystem not being provided.

> Can the generator simply run the proto generation script and then update it using the language specific formatting? 

Yes, I thought I had that enabled, but apparently not (for typescript, at least). Let me make that change.

> Can this be done without using the templating engine?

No, the templating engine makes this possible, in lieu of the manual copy-paste operations I used to do.

> There's also issues with spaces, tabs, comments etc, that always produce changes. Are we happy with the maintenance level the protoc gen plugin?

[I wrote the `protoc-gen-sdk` plugin](https://github.com/trinsic-id/protoc-gen-sdk/blob/main/action.yml), so it can have whatever support it needs. I'll make a few changes to reduce whitespace

PR #1204, opened 2022-12-16
Rely on the bundled binaries for all languages

fundthmcalculus:
Closes https://github.com/trinsic-id/server/issues/1279


.github/install-okapi-binaries-action/action.yaml
```diff
@@ -0,0 +1,11 @@
+name: "Install Okapi Binaries"
+description: "Install Okapi binaries on the github action runner"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Download Libs
+      run: |
+        curl -L -o libs.zip https://github.com/trinsic-id/okapi/releases/download/v1.6.0/libs.zip
```
janpieterz: Should the version be hard-coded? 🤔  Could we somehow curl the last release number, or retrieve it from. our local git clone?

janpieterz (approved):

.github/install-okapi-binaries-action/action.yaml
fundthmcalculus: Okapi and SDK aren't necessarily in lockstep, so we can't use the local version number. I'll see if I can curl the latest released version (I know I can, question is what's the easiest way in powershell XD)

fundthmcalculus: Hard part of this is the macos behavior is different. Let me see what I can do about that.

fundthmcalculus: Got it working - finally!

PR #1219, opened 2023-01-05
Adding new oidc param to docs

Hmac512:
This PR simple adds documentation for the new OIDC param `trinsic:mode`


Hmac512:
@tmarkovski Do we want to be more verbose than that?

geel9:
I would add a bit more info -- specifically to silent. Redirect and popup make intuitive sense, but silent demands explanation IMO.

tmarkovski (approved):

tmarkovski:
Yeah I think might be worth adding a section on what different modes are and how they impact user experience.

PR #1220, opened 2023-01-06
Update okapi binaries instructions

fundthmcalculus:


docs/java/index.md
```diff
@@ -8,7 +8,7 @@ The Trinsic Java / Kotlin SDK makes it easy to interact with the Trinsic API fro
 3. Run `gradle build` to download the required jar files. This can take some time.
 
 ## Configuration
-1. Make sure you have the okapi native binaries installed on your machine. Default installation instructions are in the [readme.md here](https://github.com/trinsic-id/okapi/)
+1. Okapi binaries are now included with the server `.jar` or the corresponding android `.aar` file. The gradle include process is essentially similar (see below).
```
janpieterz, line 11: What's the purpose of still having the gradle explanation if they are included?

janpieterz (approved):

docs/java/index.md
fundthmcalculus: You still have to custom download the `.jar` and `.aar` from github (which is not a maven repo - we've tried Maven, it creates more problems than it solves), so the gradle setup is still required. You just don't have to `winget install okapi` anymore.

PR #1231, opened 2023-01-12
Remove a bunch of old samples.

fundthmcalculus:
As part of cleaning up the samples, let's remove some we don't use that are duplicated in the `sdk-examples` repository.


samples/dotnet/Program.cs
```diff
@@ -1,111 +0,0 @@
-﻿using System;
```
janpieterz, line 1: This feels like a pretty useful sample actually for maybe all languages to have? Something simple like this showing a flow from a-z? Why would this be deleted?

janpieterz: Or are these replicated in `sdk-examples`?

fundthmcalculus: They're replicated in `sdk examples`

fundthmcalculus: Now that we have cross repository action calls, it's possible to kick off the SDK examples pipeline whenever we make an SDK release

.github/workflows/build-golang.yml
```diff
@@ -21,11 +21,11 @@ on:
 jobs:
   build-and-test-golang:
     name: Build, Test
-    runs-on: ${{ matrix.os-artifact[0] }}
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        os-artifact: [ [ubuntu-latest, linux], [windows-latest, windows], [macos-latest, macos] ]
+        os: [ ubuntu-latest, windows-latest, macos-latest ]
```
janpieterz, line 28: Why no more Linux? Feels like a different change than removing samples.

janpieterz (approved):

.github/workflows/build-golang.yml
fundthmcalculus: `ubuntu-latest` is linux. I am removing the unnecessary artifact download, not the os...

PR #1232, opened 2023-01-14
Updated concept docs with list of dids we support

fundthmcalculus:


docs/learn/concepts/dids.md
```diff
@@ -38,10 +38,12 @@ Issuance and verification of credentials against these DIDs will still function
...
-| did:ion | Distributed Ledger | ION is a sidetree/Layer-2 protocol on top of the Bitcoin blockchain. Registering a DID will incur charges if on the mainnet network. |
+| Method   | Type               | Notes                                                                                                                                                  |
+|----------|--------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------|
+| did:key  | Static             | The default DID method for all wallets. Encodes key information directly in the DID; as a result, key rotation requires a new DID.                     |
+| did:ion  | Distributed Ledger | ION is a sidetree/Layer-2 protocol on top of the Bitcoin blockchain. Registering a DID will incur charges if on the mainnet network.                   |
+| did:indy | Distributed Ledger | INDY is a Hyperledger Indy based protocol. Registering a DID will incur charges if on the main (non test) networks                                     |
```
janpieterz: Should we add which sub methods as well? Or a link to find out in our docs?

janpieterz (approved):

docs/learn/concepts/dids.md
fundthmcalculus: I would generally be leery of adding those here, because then they're going to go out of date anytime we change something. I would rather link to the appropriate enumeration and GitHub, and let them see from there or use the generated proto if I can.

fundthmcalculus: Added hyperlinks! :)

PR #1249, opened 2023-01-27
Add support for `FileManagementService`

geel9:
- Adds bindings for `FileManagementService`
- Adds bindings for `AccessManagementService` where missing


fundthmcalculus (approved):
Nice!

go/services/file_management_service.go
```diff
@@ -0,0 +1,108 @@
+package services
+
+import "github.com/trinsic-id/sdk/go/proto/services/filemanagement/v1/filemanagement"
```
fundthmcalculus, line 4: You need to import `context.Context` up here. Look at any other service for an example

geel9: hmm, I had done that before running `protoc-gen-sdk` -- this is why auto-removing unused imports on save is annoying

fundthmcalculus: That's `gofmt` for you. It's opinionated for the sake of being opinionated. It's also annoying as heck.

PR #1250, opened 2023-01-30
Implement `FileManagementService` in Typescript

geel9:
- Adds missing Typescript implementation for `FileManagementService`


web/src/FileManagementService.ts
```diff
@@ -12,8 +12,43 @@ export class FileManagementService extends ServiceBase {
...
+  /** Upload a file to Trinsic's CDN */
+  public async uploadFile(request: proto.UploadFileRequest): Promise<proto.UploadFileResponse> {
+    
+    return this.client.uploadFile(request, {
+      metadata: await this.buildMetadata(proto.UploadFileRequest.encode(request).finish())
+    });
+  }
+  /** Fetch information about a file by its ID */
+  public async getFile(request: proto.GetFileRequest): Promise<proto.GetFileResponse> {
+
```
fundthmcalculus, line 27: Did you generate this using `protoc-gen-sdk`?

fundthmcalculus (approved):

web/src/FileManagementService.ts
geel9: Yeah -- it was missed before because the Swift target doesn't exist (I'll handle that later), which caused it to give up before generating TypeScript

fundthmcalculus: Ah, good point. I'm going to be handling the `swift` target separately as part of another ticket. Want me to take that off your plate?

geel9: Sure, much appreciated

PR #1253, opened 2023-01-31
Mark Deprecated endpoints in the docs

fundthmcalculus:
* I looked for discussion of `Issue` endpoint, and everything seems to talk about issuing from templates.
* Added a warning header about deprecation

### DIFF
--- a/docs/reference/services/credential-service.md
+++ b/docs/reference/services/credential-service.md
@@ -12,6 +12,9 @@ The Credential Service exposes functionality for issuance, proof generation, ver
 
 ## Issue Credential
 
+!!! warning "Deprecation Notice"
+    This endpoint is deprecated, and will be removed on May 1, 2023. Please call [Issue Credential From Template](./credential-service.md#issue-credential-from-template).
+
 Issues a credential from a valid JSON-LD document. Issued credentials are not automatically stored in any wallet.
 
 {{ proto_sample_start() }}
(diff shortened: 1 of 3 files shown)


docs/reference/services/credential-service.md
```diff
@@ -12,6 +12,9 @@ The Credential Service exposes functionality for issuance, proof generation, ver
 
 ## Issue Credential
 
+!!! warning "Deprecation Notice"
+    This endpoint is deprecated, and will be removed on May 1, 2023. Please call [Issue Credential From Template](./credential-service.md#issue-credential-from-template).
```
janpieterz, line 16: To do this, we will need to release a new SDK for all languages. Which is long overdue anyway.

There might be some other methods or parameters that are deprecated as well. I know we should have done this before, but would you be able to add a similar note for those?
The ones that jumps to mind is the Invite one (not sure if others).

janpieterz (approved):

docs/reference/services/credential-service.md
janpieterz: The dates for all of those should be the same FYI. So we might need to rewrite some comments in `server-api`

fundthmcalculus: Agreed. I'll leave this ticket open while we decide, and make it a single update for the month.

fundthmcalculus: `ProviderService.Invite` isn't currently in the docs at all, so we're good on that front.

fundthmcalculus: Marked `ProviderService.Update` as deprecated

PR #1257, opened 2023-01-31
Name the generated proto branches intelligently

fundthmcalculus:
* Add support for `obsolete` and `experimental` tags
* Name the proto branches so we can run the action from branches as well as PRs
- [ ] Need to alias `file-management-service.dart` @geel9 maybe you can do that as [part of your PR here](https://github.com/trinsic-id/sdk/pull/1254) ?


fundthmcalculus:
This one is now redundant. Closing.

PR #1285, opened 2023-02-16
fix: Ensure dotnet SDK publishes the proper version

fundthmcalculus:
* Also ensure it has the correct number of decimal fields
* Closes https://github.com/trinsic-id/server/issues/1725


dotnet/Trinsic/ServiceBase.cs
```diff
@@ -149,8 +149,8 @@ private MetadataResponse CachedOkapiMetadata
     /// Fetches the current version of the SDK
     /// </summary>
     /// <returns></returns>
-    private string GetSdkVersion() {
+    protected internal string GetSdkVersion() {
         // This will always be 1.0.0.0 on local builds, since the version number is set on the github action during publish.
```
geel9:
```suggestion
        // This will always be 1.0.0 on local builds, since the version number is set on the github action during publish.
```

dotnet/Tests/Tests.cs
```diff
@@ -58,6 +59,11 @@ public static ServiceOptions GetTestServiceOptions()
         };
     }
 
+    [Fact]
```
geel9:
```suggestion
    [Fact(DisplayName = "SDK Version has 3 decimal places")]
```

geel9 (approved):

PR #1286, opened 2023-02-16
feat: Action to update winget .msi automatically

fundthmcalculus:


fundthmcalculus:
Need to merge so that the action registers. Then can diagnose issues.

janpieterz (approved):

PR #1299, opened 2023-02-27
Reduce docs font size

tmarkovski:


fundthmcalculus (commented):
I'm fine with the change, I'd defer to @michaeldboyd on this though.

janpieterz (approved):

geel9:
Looking at the generated preview, it looks kind of strange -- comparing it to the live version, it seems like the entire page is being squished down, not just the font size

geel9:
I guess that's actually because everything is using `em` for spacing.

Overall I guess I'm fine with it

tmarkovski:
> I guess that's actually because everything is using `em` for spacing.
> 
> Overall I guess I'm fine with it

Yeah, some of the items will need fine tuning. I think some of the font sizes we added appear overtuned now. It's not perfect, but the original size was just too large.

geel9:
I definitely agree that the font is simply way too big as-is. It's been bugging me for some time.

PR #1302, opened 2023-03-02
refactor: Deprecate webhooks

mewmba:
:crown: copied proto files from server repo: 9aa58525 refactor: Remove unused and obsolete Grant from Provider proto (#1791)
***UPDATE THIS AND PR TITLE***


janpieterz:
@fundthmcalculus  Weird, no dotnet updates, doesn't look like the webhooks methods are removed in the Typescript one at least, any ideas?

fundthmcalculus:
I'll take a look when I get home. That's odd that it didn't seem to remove it because it should have.

janpieterz:
Ouch, it's now blending changes into one branch. We should probably change this so that the mewmba changes don't override each other in the sdk, e.g. by copying branch names and doing server/xyz => sdk/xyz and sdk/xyz-generated or something?

fundthmcalculus:
> Ouch, it's now blending changes into one branch. We should probably change this so that the mewmba changes don't override each other in the sdk, e.g. by copying branch names and doing server/xyz => sdk/xyz and sdk/xyz-generated or something?

No argument there. That's been a low priority item for me, but I can definitely do that.

fundthmcalculus (approved):

PR #1328, opened 2023-03-09
spike: embedded code samples rework

fundthmcalculus:
* Injected code sample tabs are now created by a macro for consistency and simplicity
* There is an additional output on `protoc-gen-sdk` which creates all endpoints in the `docs/reference/services-generated` folder so you can copy-paste the new endpoints over. https://github.com/trinsic-id/protoc-gen-sdk/pull/55
* Working on the Dart samples of the new layout
* Required code files are added in the necessary locations, if you need/want to move them to a subfolder, we can, just talk to me so I can update the macro.


tmarkovski:
Bump. Is this work still pending or can we close this? @fundthmcalculus

fundthmcalculus:
This was work that needed to be done, but was put on hold per @janpieterz . Happy to have a further discussion about the need for it's completion, but there's still work to do.

janpieterz:
I think we can cancel the work of "all languages". But this pull request might contain some changes that are useful regardless/outside of that. Are there any?

fundthmcalculus:
> I think we can cancel the work of "all languages". But this pull request might contain some changes that are useful regardless/outside of that. Are there any?

There is, let me go though it and extract out the useful bits. The goal of this was to automate a lot of the sample generation so it's clearer where to put things.

PR #1393, opened 2023-05-08
build: Speed up the infrastructure testing by removing compiling the CLI (which takes over 3 minutes)

fundthmcalculus:


janpieterz (approved):
We should let people know, and can we confirm that this actually runs grpc 2 (feels weird to direct to a grpc-web samples directory) as well as let the team know? Everyone expects the CLI tests to run, so this is worth notifying in #engineering.

fundthmcalculus:
Agreed. I wanted to run it by you before I post it to the broader channel. I should probably rename the folder, since it's really just NPM testing. I specifically went away from the CLI because of the amount of time it takes to run `cargo build`.

tmarkovski:
> I specifically went away from the CLI because of the amount of time it takes to run cargo build.

CLI can be installed using Homebrew. The service makes bottles for Linux x86 already, so it would just download and use it without compiling.

.github/workflows/test-infrastructure.yml
```diff
@@ -19,24 +19,26 @@ on:
 
 jobs:
   test-infra:
-    name: Test Infrastructure ${{ inputs.environment }}
+    name: Test Infrastructure via HTTP2 ${{ inputs.environment }}
```
tmarkovski: Doesn't Node use grpc-web?

fundthmcalculus: It can use either. I will switch to the `brew install` CLI if that's fast enough.

fundthmcalculus:
> > I specifically went away from the CLI because of the amount of time it takes to run cargo build.
> 
> CLI can be installed using Homebrew. The service makes bottles for Linux x86 already, so it would just download and use it without compiling.

brew install actually installs via cargo, which doesn't fix the underlying problem of rust's painfully long compile times.

tmarkovski:
> brew install actually installs via cargo, which doesn't fix the underlying problem of rust's painfully long compile times.

Only if there's no bottles for the platform. Our CI creates [bottles](https://github.com/trinsic-id/homebrew-tap/blob/dc7e5979f4d6dbc15d861bf4cf93c133e324eae5/Formula/trinsic-cli.rb#L11) for [Linux x86](https://github.com/trinsic-id/homebrew-tap/releases/tag/trinsic-cli-1.9.0), so this shouldn't build the formula locally, just pull the binary.

.github/workflows/test-infrastructure.yml
tmarkovski: The CLI test is more complete, and runs trust registry tests as well. Node samples lack few use cases.
I'm not opposed to running the tests in any platform to improve speed, as long as they're covering the use cases. CLI tests are also very convenient to run locally.

fundthmcalculus:
> > brew install actually installs via cargo, which doesn't fix the underlying problem of rust's painfully long compile times.
> 
> Only if there's no bottles for the platform. Our CI creates [bottles](https://github.com/trinsic-id/homebrew-tap/blob/dc7e5979f4d6dbc15d861bf4cf93c133e324eae5/Formula/trinsic-cli.rb#L11) for [Linux x86](https://github.com/trinsic-id/homebrew-tap/releases/tag/trinsic-cli-1.9.0), so this shouldn't build the formula locally, just pull the binary.

ZOOM. 13 seconds to set up homebrew AND the cli. :rocket:

PR #1394, opened 2023-05-08
Login sample python

fundthmcalculus:
* This is an updated login sample I wrote which caught some of the issues with the new `walletAuthenticate` and `walletConfirm` workflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 ## Status
 #### Production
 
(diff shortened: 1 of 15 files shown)


janpieterz (approved):
Approved, but maybe feels better as an integration test in the server?

fundthmcalculus:
I specifically have it not as an integration test because I want to use it for customer side testing.

PR #1398, opened 2023-05-10
Fix broken tests

fundthmcalculus:
* Fix tests for removed endpoints.
* We need to do a better job of either addressing the tests when a PR is opened (and fails on CI), or notifying me that it needs to be fixed.

### DIFF
--- a/dart/example/ecosystem_example.dart
+++ b/dart/example/ecosystem_example.dart
@@ -13,7 +13,7 @@ Future runEcosystemDemo() async {
   // createEcosystem() {
   var actualCreate = await trinsic.provider().createEcosystem(
       request: CreateEcosystemRequest(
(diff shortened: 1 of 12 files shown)


geel9 (approved):

janpieterz:
@fundthmcalculus if they fail on CI how does it get merged?

fundthmcalculus:
That was more in common about people not updating when they just close the mewmba PRs. In this case, I told Ciara to merge while I wrote the tests so that she wasn't blocked on OIDC work.

janpieterz:
Ah I see - aren't we compiling these samples and tests? The CI should find this no?

fundthmcalculus:
It does. The problem has been that most people see the CI failures and just close the PR rather than actually merging any updates.

PR #1425, opened 2023-06-07
CLI Release / windows installation docs changes

geel9:
- Reverts the version in `cli.yaml` to `1.0.0` as release actions depend on this constant
- Sets the asset value attached to the GitHub Release for Windows and Debian CLI Installers to include `cli` in the name
- Updates the docs to not recommend installing CLI from `winget` as we do not seem to be pushing to it anymore


fundthmcalculus (approved):
Thanks! winget doesn't support dependencies installation.

PR #1462, opened 2023-08-10
build: Switch to vite packaging

fundthmcalculus:
* While certainly faster (and smaller), the output is hard or impossible to debug.
* Added source maps and type definitions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -408,3 +408,4 @@ dart/test/dart_test.dart.vm.json
 samples/grpc-web/build/
 samples/grpc-web/lib/
 samples/grpc-web/coverage/
+web/dist/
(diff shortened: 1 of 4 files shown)


tmarkovski:
> * While certainly faster (and smaller), the output is hard or impossible to debug.

Minified JS isn't supposed to be debuggable code. Including map files will help with that.
https://vitejs.dev/config/build-options.html#build-sourcemap

> * We would still need to run `tsc` to get all the type definition information, which seems ill-advised (paying the performance penalty twice).

We can (and should) provide type definitions with a minified JS, so developers can get the editor/compiler support for the SDK, in addition to map files, so the actual JS can be debugged.

https://github.com/qmhc/vite-plugin-dts
https://stackoverflow.com/a/73123436/379550

fundthmcalculus:
THANK YOU! Somehow I couldn't find the source map output on `vite`. Thanks for the links!

lucasamonrc (approved):

PR #1488, opened 2023-09-04
Docs improvements

chiarapaskulin:


chiarapaskulin:
Some changes included here don't belong to this PR. They will disappear once we merge [#2730](https://github.com/trinsic-id/server/pull/2730).

chiarapaskulin:
Not ready for review yet. Just marked it as ready for review to see mewmba's docs previews :) (thanks steven @fundthmcalculus for those! <3 )

tmarkovski:
Why isn't the docs preview action firing?

docs/examples/1-issue-direct-send.md
```diff
@@ -79,47 +79,47 @@ const issueResponse = await trinsic.credential().issueFromTemplate({
 console.dir(JSON.parse(issueResponse.documentJson), { depth: null, });
 ```
 
-Let's look at some of the input fields for the issue from template request.
-
-:   `templateId` is a required field and refers to the template ID that we want to use.
```
tmarkovski: Any reason why you chose to use `<li>` as opposed to the definition indented list that comes with MkDocs?

docs/examples/1-issue-direct-send.md
```diff
@@ -79,47 +79,47 @@ const issueResponse = await trinsic.credential().issueFromTemplate({
...
+        ```
+
+    <li>`expirationDate` is optional and specifies the validity period of the issed credential. If specified, this field must be in ISO 8601 format.</li>
+    <li>`saveCopy` is optional and will save a copy of the issued credential in the issuer wallet that will be marked as `IssuedVerifiableCredential`. This copy will not contain any cryptographic data or signatures and it's purpose is to allow the issuers to track the issued credentials and their metadata. This metadata is required if the issuer wants to change the status of the credential and revoke it's validity. This is controlled using the data found in the `credentialStatus` field.</li>
+<ul>
 
 ### Send the credential
 
-Finally, we can just send the credentil to the user's wallet by specifying the email address of the user.
+Finally, we can just send the credentil to the user's wallet by specifying its email address. If you want to send an email notification that a credential has been sent to the user's wallet, just set `sendNotification` as true.
```
tmarkovski:
```suggestion
Finally, we can just send the credential to the user's wallet by specifying its email address. If you want to send an email notification that a credential has been sent to the user's wallet, just set `sendNotification` to `true`.
```

fundthmcalculus:
> Why isn't the docs preview action firing?

I need to get the new mkdocs-material token in.

lucasamonrc (commented):
I noticed the same things that @tmarkovski pointed out so I will not give a explicit approval for now. But once those comments are addressed I'll approve it.

fundthmcalculus:
@chiarapaskulin your demo docs site is up!

docs/examples/1-issue-direct-send.md
```diff
@@ -29,7 +29,7 @@ const trinsic = new TrinsicService({
 });
 ```
 
-You can obtain auth token from the [Dashboard <small>:material-open-in-new:</small>](https://dashboard.trinsic.id){targer=_blank} by clicking on the View Auth Token in the Home page.
+(You can obtain Auth Token from the [Dashboard <small>:material-open-in-new:</small>](https://dashboard.trinsic.id){targer=_blank} by clicking on the View Auth Token button in the Home page.)
```
tmarkovski:
```suggestion
(You can obtain an Auth Token from the [Dashboard <small>:material-open-in-new:</small>](https://dashboard.trinsic.id){targer=_blank} by clicking on the View Auth Token button in the Home page.)
```

tmarkovski (approved):

docs/examples/1-issue-direct-send.md
chiarapaskulin: Oh, I used it to have the dots in the list, like this:
image
But I found out I can use " - " (MkDocs docmark) to have the dots! :)
Thanks!

mkdocs.yml
```diff
@@ -55,7 +55,6 @@ nav:
       - Wallet API: reference/services/wallet-service.md
     - Other Services:
       - OpenID Credential Exchange: reference/other/openid.md
-  - Examples: examples/index.md
```
lucasamonrc: Please, restore this line. We need to have examples in the navigation.

lucasamonrc (changes requested):

mkdocs.yml
chiarapaskulin: Ops, that's an old change I reverted but somehow is still here... Thanks!

lucasamonrc (approved):

lucasamonrc (approved):

PR #1493, opened 2023-09-08
[typescript] Add ConnectClient and separate browser/node SDK entry

tmarkovski:
- Separates the SDK so that there's different entry for browsers
- Adds bundled JS file in `dist`
- Add ConnectClient available only for browsers
- Wrap OIDC client in as a single SDK call in ConnectClient

No implementation of ConnectClient for IDV, this just brings in the required work.


fundthmcalculus (commented):
Can you separate the `prettier` format work into a separate PR, it makes this one much harder to review.

tmarkovski:
> Can you separate the `prettier` format work into a separate PR, it makes this one much harder to review.

How I do that without making pipelines fail?

It's not that many files, if you filter without "whitespace", just ignore everything under `proto`

PR #1496, opened 2023-09-11
Use webpack for bundling. Fix popup callback.

tmarkovski:
For some reason, I couldn't get vite to properly bundle without erroring at runtime. Webpack seemed to work, so going with that for now.
`vite` assets are still configured if anyone wants to take a stab at it.


fundthmcalculus:
@Hmac512 got any thoughts on this. I know we're trying to switch to `vite`.

tmarkovski:
FYI, I published `-pre` release version of this, so I can work on the examples and docs.

PR #1516, opened 2023-09-14
Adding redirect uri to connect client

Hmac512:


tmarkovski (approved):

web/src/ConnectClient.ts
```diff
@@ -240,18 +240,23 @@ import MicroModal from "micromodal";
...
+    constructor(
+        connectUrl: string = "https://connect.trinsic.cloud",
+        redirectUri?: string
+    ) {
         this.baseUrl = connectUrl;
+        this.redirectUri = redirectUri;
 
         this.oidcConfig = {
             authority: this.baseUrl,
             client_id: "http://localhost:8080/",
```
geel9, line 258: I know this isn't part of the PR, but this seems... wrong for something we're shipping to customers. Is it possible to just elide this parameter if not provided?

tmarkovski: No, as it's required for OIDC. We can change it to something else.

PR #1524, opened 2023-09-21
fix: Update ConnectClient.ts to prevent scrollbars in iframe

lucasamonrc:


web/src/ConnectClient.ts
```diff
@@ -323,7 +323,7 @@ export class ConnectClient {
         modalContainer.ariaModal = "true";
 
         modalContainer.className = mobileDetect.isDesktop
-            ? "modal__container h-[600px] w-[400px]"
+            ? "modal__container h-[600px] w-[400px] overflow-hidden"
```
Hmac512:
```suggestion
            ? "modal__container h-[600px] w-[400px] lock-bg"
```

Hmac512 (changes requested):

PR #1534, opened 2023-09-28
Fixing popup staying open

Hmac512:


lucasamonrc (approved):

tmarkovski:
Is this required? We don't use the "oidc-client" source anymore in the callbacks.

PR #1600, opened 2024-02-16
Refactor/remove old samples

fundthmcalculus:
* Moves `grpc-web` infra testing to the SDK itself, which is actually kept up to date.
* Removes old samples that aren't useful
* Moves the ecosystem-migration python sample over to `/examples`


fundthmcalculus:
Will fix golang in a different PR.

PR #1667, opened 2024-06-13
fix: Add `ConnectClient` to new SDK

fundthmcalculus:


janpieterz:
We probably want to rename it, and this should be in the client sdk. Let's discuss how that would work - eg typescript runs on backend, web, react native, how do we blend those.

